PR #16, opened 2019-09-10
Remove comment from make-help

ayushkovskiy:


anayden:
I think we should keep the credit. Also, there's a comment that contains a minor fix for this command line in case there're includes used: http://disq.us/p/1637kir

ayushkovskiy:
fixed, thanks

PR #17, opened 2019-09-10
Several minor fixes

### LINKED ISSUE #4: minor: rename underscored commands to dash-separated commands
In neuro, we use `neuro job port-forward`. Let's use same `make upload-code` instead of `make upload_code`?

anayden:
1. Add make target sorting by name in `make help`
2. Replace understore with dash in target names (fixes #4)
3. Remove empty line in the Makefile
4. Add `--browse` to `make tensorboard` and `make filebrowser` targets

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
-neuromation-template
+Neuro Platform Project Template
 ====================
 
-Base template for Neuromation-based projects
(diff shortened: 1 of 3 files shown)


ayushkovskiy:
LGTM without sorting

PR #21, opened 2019-09-11
Set default package name to `modules`

### LINKED ISSUE #15: Set default package name to `modules`
When set to `modules` the user will be able to put Python files into '/modules` and import them as 

```python
from modules import train`
```

However they can override it, typing a short project name, for example, and import it as

```python
from my_project import train
```

anayden:
Closes #15

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -1,6 +1,6 @@
 {
     "full_name": "Your name",
-    "email": "Your email address (eq. you@example.com)",
+    "email": "Your email address (e.g. you@example.com)",
     "project_name": "Name of the project",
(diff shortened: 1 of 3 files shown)


cookiecutter.json
```diff
@@ -1,11 +1,9 @@
 {
     "full_name": "Your name",
-    "email": "Your address email (eq. you@example.com)",
-    "github_username": "Your github username",
+    "email": "Your email address (eq. you@example.com)",
```
ayushkovskiy: by the way: is `eq` a standard abbreviation for `e.g.`?

anayden: hm…I never noticed it says `eq.` here. I've no idea what that means :)

PR #44, opened 2019-09-16
Fix jupyter issues

### LINKED ISSUE #27: Add add project name suffix to job names
Tried running `make jupyter` while another job with the same was running and got an error. It was easy for me to identify the root cause, but might not be the case for those who is not familiar with `neuro`. I was in the situation when I had job named `jupyter` running already and wanted to keep it.

### LINKED ISSUE #28: Add dependencies for jupyter target
it needs to depend on `upload_code` and `upload_notebooks`

anayden:
Closes #27
Closes #28

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -115,7 +115,7 @@ connect-training:  ### Execute shell to the training job
 	neuro exec --no-key-check $(TRAINING_NAME) bash
 
 .PHONY: jupyter
-jupyter:  upload_code upload_notebooks ### Run jupyter job
+jupyter:  upload-code upload-notebooks ### Run jupyter job
 	neuro run \
 		--name $(JUPYTER_NAME) \
 		--preset gpu-small \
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -115,7 +115,7 @@ connect-training:  ### Execute shell to the training job
 	neuro exec --no-key-check $(TRAINING_NAME) bash
 
 .PHONY: jupyter
-jupyter:  ### Run jupyter job
+jupyter:  upload_code upload_notebooks ### Run jupyter job
```
mariyadavydova: The targets are already renamed. They use `-` instead of `_`.

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -21,10 +21,10 @@ REQUIREMENTS_APT_ENV?=$(PROJECT_PATH_ENV)/$(REQUIREMENTS_APT)
 RESULTS_PATH_ENV?=$(PROJECT_PATH_ENV)/$(RESULTS_PATH)
 
 SETUP_NAME?=setup
-TRAINING_NAME?=training
-JUPYTER_NAME?=jupyter
-TENSORBOARD_NAME?=tensorboard
-FILEBROWSER_NAME?=filebrowser
+TRAINING_NAME?=training-{{cookiecutter.project_slug}}
```
mariyadavydova: I think we may use the same idea for setup, it may take some time as well.

{{cookiecutter.project_slug}}/Makefile
anayden: Blind copy-paste from issues = evil :(
Thanks for this catch!

{{cookiecutter.project_slug}}/Makefile
anayden: ok, sounds good

PR #55, opened 2019-09-23
Add testing environments

ayushkovskiy:
now we can choose the environment to run e2e tests against: `make test_e2e_dev` (any branch) or `make test_e2e_staging` (only `master`)


Makefile
```diff
@@ -29,6 +29,12 @@ test_unit:
 	stat ../test-project
 	python -m doctest tests/e2e/conftest.py
 
-.PHONY: test_e2e
-test_e2e:
-	pytest -v -s --reruns=2 tests/e2e
+.PHONY: test_e2e_dev
+test_e2e_dev:
+	export TRAINING_MACHINE_TYPE=cpu-small
+	pytest -v -s --environment=dev tests/e2e
```
anayden: Should we have `reruns=2` here as well?

ayushkovskiy: I'm not sure: the logic encoded here was to make staging tests more robust (`--reruns=2`) and to make dev tests more informative (`--reruns=0` for detecting all kinds of flakes instantly). So I propose to stick with this scheme for the first time, and later adjust values.

anayden: I see, that makes sense now

anayden:
LGTM when green :)

tests/e2e/conftest.py
```diff
@@ -156,8 +195,8 @@ def generate_empty_project(run_cookiecutter: None) -> None:
 
 @pytest.fixture(scope="session", autouse=True)
 def pip_install_neuromation() -> None:
-    run("pip install -U neuromation")
-    assert "Name: neuromation" in run("pip show neuromation")
+    run("pip install -U neuromation", verbose=False)
```
ayushkovskiy: forgotten fixes

Makefile
```diff
@@ -29,6 +29,10 @@ test_unit:
 	stat ../test-project
 	python -m doctest tests/e2e/conftest.py
 
-.PHONY: test_e2e
-test_e2e:
-	pytest -v -s --reruns=2 tests/e2e
+.PHONY: test_e2e_dev
+test_e2e_dev:
+	TRAINING_MACHINE_TYPE=cpu-small pytest -v -s -x --environment=dev --tb=line tests/e2e
```
ayushkovskiy: Note, I added `-x` option for both dev and staging (aborts execution once a test failed)

PR #58, opened 2019-09-23
Add data/ to .gitignore

### LINKED ISSUE #53: `data` folder content should be ignored
It should contain `.gitignore` which ignores everything but this file.

anayden:
Closes #53

### DIFF
--- a/{{cookiecutter.project_slug}}/.gitignore
+++ b/{{cookiecutter.project_slug}}/.gitignore
@@ -124,3 +124,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# data/ folder
+data/**/*


{{cookiecutter.project_slug}}/.gitignore
```diff
@@ -124,3 +124,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# data/ folder
+data/**/*
```
ayushkovskiy: by the way: maybe it's better to prepend all dirs lying in the root directory with slash: `/data/**/*`, `/venv/`, ...?

anayden: Looks like, for instance, [gitignore.io](https://www.gitignore.io/api/python) doesn't do it for most folders. To be honest, I'm not sure what the convention is here, but it seems the default is not to use leading slash.

What do you think?

mariyadavydova: And another question: why this solution is better than the one I suggested in the template initially (https://github.com/neuromation/platform-benchmark/blob/neuro-template/data/.gitignore)?

anayden: >  if I have a file prices.csv in data folder, this file WILL be under Git?

No, it will be ignored.

> why this solution is better than the one I suggested in the template initially

Your solution splits ignored files across multiple `.gitignore`'s, whereas this one keeps all of them in one place. I think it's very counterintuitive to look for `.gitignore` in a subfolder of your project. I would never guess to do that and I would struggle to understand why my files are ignored by git even though `data/` is not in the "main" `.gitignore`

mariyadavydova: OK, makes sense.

PR #59, opened 2019-09-23
Improve test regexps

ayushkovskiy:


anayden:
I guess logs should be gitignored

ayushkovskiy:
> I guess logs should be gitignored

sorry log is a wrong file came from another PR. Already removed

ayushkovskiy:
closed as the older PR has changed completely from its original version.

PR #60, opened 2019-09-23
Generate unique project names for each test run

ayushkovskiy:
otherwise any 2 builds interfere and break each other

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,20 +1,18 @@
-import logging
 import os
 import re
 import shutil
-import signal
 import sys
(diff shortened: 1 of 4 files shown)


tests/e2e/conftest.py
```diff
@@ -198,48 +190,6 @@ def neuro_login(pip_install_neuromation: None) -> t.Iterator[None]:
             log.info("Deleted")
 
 
-# == generic helpers ==
```
ayushkovskiy: these three methods need to be moved to `utils.py` so that we have access to them from `configuration.py`

PR #64, opened 2019-09-23
Improve testing patterns

ayushkovskiy:
Use `[^\n]` instead of `.` in stop-patterns so that more information is caught in messages like `ERROR: ...\r\n`


ayushkovskiy:
not relevant anymore

PR #65, opened 2019-09-23
Fix race with checking `neuro ps` in tests

ayushkovskiy:
When multiple builds run in parallel, they conflict in the output of `neuro ps`


tests/e2e/conftest.py
```diff
@@ -458,11 +458,13 @@ def neuro_rm_dir(
...
+        f"neuro status {job_id}",
         timeout_s=timeout,
-        debug=True,
+        debug=False,
         stop_patterns=DEFAULT_NEURO_ERROR_PATTERNS,
     )
-    return set(out.split())
+    search = re.search(r"Status: (\w+)", out)
+    assert search, f"not found job status in output: `{out}`"
+    return search.group(1)
```
anayden: I feel like returning `True/False` would be more convenient. It feels weird to compare return value of `neuro_status` with `succeeded` — what does it imply to a person reading the code?

ayushkovskiy: done thanks!

PR #72, opened 2019-09-24
Mute neuro-kill errors

ayushkovskiy:
Fixes https://github.com/neuromation/cookiecutter-neuro-project/issues/71

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -55,7 +55,7 @@ help:
 
 .PHONY: setup
 setup: ### Setup remote environment
-	neuro kill $(SETUP_NAME)
+	neuro kill $(SETUP_NAME) >/dev/null 2>&1
 	neuro run \
 		--name $(SETUP_NAME) \
 		--preset cpu-small \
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -55,7 +55,7 @@ help:
 
 .PHONY: setup
 setup: ### Setup remote environment
-	neuro kill $(SETUP_NAME)
+	neuro kill $(SETUP_NAME) >/dev/null 2>&1
```
anayden: I would extract `neuro kill >/dev/null 2>&1` to a variable and reuse everywhere. Something like `$(NEURO_KILL) $(SETUP_NAME)`, just like we do with `neuro cp`

mariyadavydova: @anayden, is it correct that
```
neuro kill setup-project >/dev/null 2>&1
```
and
```
neuro kill >/dev/null 2>&1 setup-project
```
work the same?

anayden: As far as I know — yes

ayushkovskiy: > I would extract neuro kill >/dev/null 2>&1 to a variable and reuse everywhere.

Is this kind of meta-programming possible in the syntaxes of Bash and Makefile?

mariyadavydova: I would rather call it Bash magic... This won't work on Windows, BTW.

">/dev/null" redirects all output to nowhere 

"2>&1" redirects stderr to stdout

Alex Davydow recommends putting these instructions before "neuro kill".

ayushkovskiy: I know, but I something like this won't work:
```
redirect=">/dev/null 2>&1"
neuro kill shomething $redirect
$redirect neuro kill something
```
@anayden what was your idea of shortcutting this syntax?

ayushkovskiy: > Alex Davydow recommends putting these instructions before "neuro kill".

Why? I have never seen it, and, to be honest, `>/dev/null 2>&1 neuro kill $(SETUP_NAME)` looks much more scary for me than `neuro kill $(SETUP_NAME) >/dev/null 2>&1`.

> I would extract neuro kill >/dev/null 2>&1 to a variable and reuse everywhere.

I analyzed the issue and came to that this has to be done only once for `make setup` -- so this is not a problem, even though Makefile supports the following syntax: 
```
MUTE:=">/dev/null 2>&1"
setup:
    ${MUTE} neuro kill job-name
```

PR #77, opened 2019-09-24
Do `make clean` only after

ayushkovskiy:
test `test_make_run_something_useful` also uses the `data/` that we cleaned up from storage at the end of test `test_make_upload_clean_data`. Here we extract tests like `test_make_clean_data` to separate tests running at the end.


.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           name: clean up running jobs
           command: |
             . venv/bin/activate
-            python cleanup_jobs.py
+            python cleanup_jobs.py || echo "Cleanup failed"
```
ayushkovskiy: temp change: will be removed in https://github.com/neuromation/cookiecutter-neuro-project/issues/78

tests/e2e/configuration.py
```diff
@@ -6,7 +6,7 @@
 TIMEOUT_MAKE_UPLOAD_NOTEBOOKS = TIMEOUT_MAKE_DOWNLOAD_NOTEBOOKS = 5
 TIMEOUT_MAKE_CLEAN_NOTEBOOKS = 5
 
-TIMEOUT_NEURO_LOGIN = 5
+TIMEOUT_NEURO_LOGIN = 15
```
ayushkovskiy: also, until our timeouts are stabilized, we'll adjust them many times in future PRs

PR #97, opened 2019-09-27
Install users code so that it works from the notebooks

ayushkovskiy:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/29, and also closes https://github.com/neuromation/cookiecutter-neuro-project/issues/94, closes https://github.com/neuromation/cookiecutter-neuro-project/issues/95


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -60,14 +62,16 @@ setup: ### Setup remote environment
 		--name $(SETUP_NAME) \
 		--preset cpu-small \
 		--detach \
-		--volume $(PROJECT_PATH_STORAGE):$(PROJECT_PATH_ENV):ro \
+		--volume $(PROJECT_PATH_STORAGE):$(PROJECT_PATH_ENV):rw \
```
ayushkovskiy: `pip install -e .` writes files to the project directory.

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -143,7 +147,7 @@ jupyter: upload-code upload-notebooks ### Run a job with Jupyter Notebook and op
 		--volume $(RESULTS_PATH_STORAGE):$(RESULTS_PATH_ENV):rw \
 		--env PLATFORMAPI_SERVICE_HOST="." \
 		$(CUSTOM_ENV_NAME) \
-		'jupyter notebook --no-browser --ip=0.0.0.0 --allow-root --NotebookApp.token= --notebook-dir=$(NOTEBOOKS_PATH_ENV)'
+		'jupyter notebook --no-browser --ip=0.0.0.0 --allow-root --NotebookApp.token= --notebook-dir=$(PROJECT_PATH_ENV)'
```
ayushkovskiy: we should start notebook for the project root directory so that the imports like `from modules import *` work

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -60,14 +62,16 @@ setup: ### Setup remote environment
...
 		--preset cpu-small \
 		--detach \
-		--volume $(PROJECT_PATH_STORAGE):$(PROJECT_PATH_ENV):ro \
+		--volume $(PROJECT_PATH_STORAGE):$(PROJECT_PATH_ENV):rw \
 		--env PLATFORMAPI_SERVICE_HOST="." \
 		$(BASE_ENV_NAME) \
 		'sleep 1h'
 	neuro cp $(REQUIREMENTS_APT) $(REQUIREMENTS_APT_STORAGE)
 	neuro cp $(REQUIREMENTS_PIP) $(REQUIREMENTS_PIP_STORAGE)
+	for file in $(PROJECT_PYTHON_FILES); do neuro cp ./$$file $(PROJECT_PATH_STORAGE)/$$file; done
```
ayushkovskiy: copy `setup.py` and other files

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -60,14 +62,16 @@ setup: ### Setup remote environment
...
 	neuro exec --no-key-check $(SETUP_NAME) "bash -c 'export DEBIAN_FRONTEND=noninteractive && $(APT_COMMAND) update && cat $(REQUIREMENTS_APT_ENV) | xargs -I % $(APT_COMMAND) install --no-install-recommends % && $(APT_COMMAND) clean && $(APT_COMMAND) autoremove && rm -rf /var/lib/apt/lists/*'"
 	neuro exec --no-key-check $(SETUP_NAME) "bash -c '$(PIP_INSTALL_COMMAND) -r $(REQUIREMENTS_PIP_ENV)'"
+	neuro exec --no-key-check $(SETUP_NAME) "bash -c '$(PIP_INSTALL_COMMAND) -e $(PROJECT_PATH_ENV)'"
```
ayushkovskiy: install user's code (for issue https://github.com/neuromation/cookiecutter-neuro-project/issues/29)

asvetlov: I quite don't understand why do you need to install the package at all?
Python notebook doesn't require it.
Other tools should work fine imho.

ayushkovskiy: otherwise `from modules import foo` does not work in the jupyter notebook

out.asciidoc
```diff
@@ -0,0 +1,4 @@
+----
```
mariyadavydova: What is this file for?

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -4,6 +4,8 @@ NOTEBOOKS_PATH?=notebooks
 REQUIREMENTS_PIP?=requirements.txt
 REQUIREMENTS_APT?=apt.txt
 RESULTS_PATH?=results
+PROJECT_PYTHON_FILES=setup.py setup.cfg README.md LICENSE
```
mariyadavydova: Why do you need to copy README.md and LICENCE? I can't imagine a use-case when having any of these two in the storage may be useful.

{{cookiecutter.project_slug}}/Makefile
mariyadavydova: Our current setup is too flexible for such a change. You may theoretically put your notebooks in some different location that '$(PROJECT_PATH_ENV)/notebooks'; running Jupyter like this will make them unavailable.

We should either find a way to make modules accessible without this change, or make the template structure less flexible.

out.asciidoc
ayushkovskiy: Thank you for pointing out on it: This file should be auto-generated by test `test_make_setup`. I'll add it to .gitignore.

{{cookiecutter.project_slug}}/Makefile
ayushkovskiy: That's why this PR changes notebooks' working directory to the project root

{{cookiecutter.project_slug}}/Makefile
ayushkovskiy: fixed not to use `setup.py`, thanks

out.asciidoc
ayushkovskiy: changed to use a temp file

{{cookiecutter.project_slug}}/Makefile
ayushkovskiy: My logic was that the template should create an environment as much close to the local environment as possible. However, I agree, README.md and LICENSE can be deleted.

{{cookiecutter.project_slug}}/Makefile
ayushkovskiy: not the case anymore. Still, I think, there's no point of mounting project root as `ro` during installation -- Let's not mount it then.

{{cookiecutter.project_slug}}/notebooks/Untitled.ipynb
```diff
@@ -0,0 +1,49 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# change working directory to the project root:\n",
+    "%cd /project\n",
```
ayushkovskiy: Note: here we explicitly change notebook's working directory to the project root, because the notebook still works in directory `{project_root}/notebooks` even if we started it with `--notebook-dir=$(PROJECT_PATH_ENV)`:

Firstly, the jupyter does start in the project root (`/project`), but then, the IPython kernel changes its directory to `/project/notebooks`:
```
[I 11:57:22.720 NotebookApp] Serving notebooks from local directory: /project
...
[D 11:57:37.078 NotebookApp] Kernel args: {'kernel_name': 'python3', 'cwd': '/project/notebooks'}
```
I agree that `%cd /project` in the top of jupyter notebooks looks bad, but I could not find the place where the kernel `python3` takes its `cwd` in the log above.

{{cookiecutter.project_slug}}/Makefile
ayushkovskiy: As @asvetlov has pointed out, we should not create a Python package for the users (using `setup.py`, see PR https://github.com/neuromation/cookiecutter-neuro-project/pull/107) because Python packaging is for *distributing* Python code, but in order to run it locally we can simply use `import modules` while running from the project root directory. As far as I understood, `pip install -e .` makes unnecessary changes to the environment (also, our `setup.py` was incorrect). So we should better enforce our notebooks to have working directory = project root. See also comment https://github.com/neuromation/cookiecutter-neuro-project/pull/97/files#r329563695

ayushkovskiy: > We should either find a way to make modules accessible without this change,

Please note that the older setting `--notebook-dir=$(NOTEBOOKS_PATH_ENV)` was useless as well, since there's no point for us to run jupyter in the `/project/notebooks` working directory

{{cookiecutter.project_slug}}/notebooks/Untitled.ipynb
ayushkovskiy: Unfortunately, I could not find the solution for this problem (asked in jupyter project: https://github.com/jupyter/notebook/issues/4924). Here (https://stackoverflow.com/questions/15680463/change-ipython-jupyter-notebook-working-directory) we are also advised to use `%cd`.

setup.cfg
```diff
@@ -1,9 +1,10 @@
 [flake8]
-exclude = .git,.env,venv,__pycache__,.eggs
+exclude = .git,.env,venv,__pycache__,.eggs,{{cookiecutter.project_slug}}
```
ayushkovskiy: hmm, I am not sure: do we really need to exclude user's code directory from linters?

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -65,6 +67,7 @@ setup: ### Setup remote environment
 		--env PLATFORMAPI_SERVICE_HOST="." \
 		$(BASE_ENV_NAME) \
 		'sleep 1h'
+	for file in $(PROJECT_PYTHON_FILES); do neuro cp ./$$file $(PROJECT_PATH_STORAGE)/$$file; done
```
ayushkovskiy: for making linters work from inside the job, not really necessary for the functionality introduced in htis PR

setup.cfg
ayushkovskiy: Achso, this is our setup.cfg, not the template's. Keeping changes.

PR #107, opened 2019-09-30
Delete setup.py

ayushkovskiy:
No need to form a [wrong] Python-package


ayushkovskiy:
See discussion https://github.com/neuromation/cookiecutter-neuro-project/pull/97/files#r329565745 and https://github.com/neuromation/cookiecutter-neuro-project/pull/97/files#r329563695

@mariyadavydova I'll wait for your approval as well

PR #113, opened 2019-10-01
Remove --detach for most make targets

### LINKED ISSUE #102: Remove `--detach` from `make jupyter` (and others)
it's very useful to see job's logs, even together with `--browse`

anayden:
Closes #102 

I did not remove `--detach` from `make setup` though. Let me know if you think that's necessary.

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -295,15 +295,25 @@ def run_once(
     `RuntimeError` is raised. Use `verbose=True` to print useful information
     to log (also to dump all child process' output to the handler defined
     in `PEXPECT_DEBUG_OUTPUT_LOGFILE`).
-    >>> # Check expected-outputs:
-    >>> run("echo 1 2 3", expect_patterns=['1', '3'], verbose=False)
+    >>> # Expect the first and the last output:
+    >>> run_once("echo 1 2 3", expect_patterns=[r'1 \d+', '3'], verbose=False)
+    '1 2 3'
+    >>> # Abort once all the patterns have matched:
+    >>> run_once("bash -c 'echo 1 2 3 && sleep infinity'",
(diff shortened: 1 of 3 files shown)


ayushkovskiy:
> I did not remove --detach from make setup though. Let me know if you think that's necessary.

sure, there are commands going after `neuro job run`

ayushkovskiy:
Tests on `make jupyter` fail. Either the job fails and we try to `curl` the `succeeded` job, or the process that submits the job `run(...)` does not return because the job is not `--detach`, so the test fails with timeout. We need to enhance helpers to return once all the patterns have matched. I'll do this.

tests/e2e/conftest.py
```diff
@@ -339,7 +353,6 @@ def run_once(
                 chunk = _get_chunk(child)
                 output += chunk
     finally:
-        output += _read_till_end(child)
```
ayushkovskiy: this blocked until the process finished. Added a test with `sleep infinity` as the last command.

tests/e2e/conftest.py
```diff
@@ -354,36 +367,6 @@ def _get_chunk(child: pexpect.pty_spawn.spawn) -> str:
     return chunk
 
 
-def _read_till_end(child: pexpect.spawn) -> str:
```
ayushkovskiy: not needed code anymore

tests/e2e/test_e2e_template.py
```diff
@@ -258,6 +259,10 @@ def test_make_run_something_useful(target: str, path: str, timeout_run: int) ->
             )
         wait_job_change_status_to(job_id, "succeeded")
 
+    except Exception:
```
ayushkovskiy: to see the error immediately

PR #122, opened 2019-10-08
Validate project slug via pre-hook

### LINKED ISSUE #68: Handle invalid project names correctly
We need to handle correctly projects with names that contain characters invalid for Python module name (`Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability.` https://www.python.org/dev/peps/pep-0008/). 

Currently, we allow some invalid names (for example, we allow names with dashes), but some features may be broken in future (for example, `make lint` will fail with `detection-kit is not a valid Python package name`).

### LINKED ISSUE #101: Restrict project name: Too long job names possible
job name max length: 40, the template will be broken if the project name is long (`jupyter-{project-name}`)

anayden:
Closes #68 
Closes #101

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -30,7 +30,7 @@ TENSORBOARD_NAME?=tensorboard-$(PROJECT_POSTFIX)
 FILEBROWSER_NAME?=filebrowser-$(PROJECT_POSTFIX)
 
 BASE_ENV_NAME?=neuromation/base
-CUSTOM_ENV_NAME?=image:neuromation-{{cookiecutter.project_slug.replace('_', '-')}}
+CUSTOM_ENV_NAME?=image:neuromation-$(PROJECT_POSTFIX)
 TRAINING_MACHINE_TYPE?=gpu-small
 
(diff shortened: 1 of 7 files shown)


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -22,14 +22,14 @@ RESULTS_PATH_ENV?=$(PROJECT_PATH_ENV)/$(RESULTS_PATH)
 
 NEURO_CP=neuro cp --recursive --update --no-target-directory
 
-SETUP_NAME?=setup-{{cookiecutter.project_slug}}
-TRAINING_NAME?=training-{{cookiecutter.project_slug}}
-JUPYTER_NAME?=jupyter-{{cookiecutter.project_slug}}
-TENSORBOARD_NAME?=tensorboard-{{cookiecutter.project_slug}}
-FILEBROWSER_NAME?=filebrowser-{{cookiecutter.project_slug}}
+SETUP_NAME?=setup-{{cookiecutter.project_slug.replace('_', '-')}}
```
ayushkovskiy: let's move this value as postfix `{{cookiecutter.project_slug.replace('_', '-')}}`

PR #127, opened 2019-10-08
Retry only usually-failing tests 

ayushkovskiy:
as discussed with @asvetlov , we should retry only tests that usually fail. For now, it's `make setup`. 

NOTE: during `make setup`, the most unstable part is `neuro save`. Firstly, I was planning to split `make setup` into few separate targets so that to test them individually (and re-run only `neuro save` part), but this is non-fair testing. So better test exact `make setup` and accept the overhead.

TODO: change the retry strategy: retry only when certain patterns found and break otherwise

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,8 +29,8 @@ test_unit:
 
 .PHONY: test_e2e_dev
 test_e2e_dev:
-	TRAINING_MACHINE_TYPE=cpu-small pytest -s --environment=dev --tb=line --reruns=2 tests/e2e
+	TRAINING_MACHINE_TYPE=cpu-small pytest -s --environment=dev --tb=short tests/e2e
 
 .PHONY: test_e2e_staging
 test_e2e_staging:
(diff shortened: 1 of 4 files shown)


tests/e2e/conftest.py
```diff
@@ -75,8 +75,14 @@
     else sys.stdout
 )
 
-# note: ERROR, being the most general error, must go the last
-DEFAULT_NEURO_ERROR_PATTERNS = ("404: Not Found", "Status: failed", r"ERROR[^:]*: .+")
+# note: ERROR, being the most general error, should go the last
+DEFAULT_NEURO_ERROR_PATTERNS = (
+    r"404: Not Found",
+    r"Status:[^\n]+failed",
+    r"Docker API error: Failed to save job [^\n]+502 Bad Gateway",
```
ayushkovskiy: the most common flakiness pattern

ayushkovskiy:
not urgent PR, for now we can leave with the strategy to retry all failed tests

tests/e2e/conftest.py
ayushkovskiy: just to print better errors

PR #133, opened 2019-10-13
Increase timeouts to work better on dev

ayushkovskiy:


ayushkovskiy:
so far, no need to increase timeouts, tests started to work against `dev` cluster after a week

PR #136, opened 2019-10-15
Remove version numbers from requirements.txt

### LINKED ISSUE #117: requirements.txt: do not fix versions of base pip dependencies
they are already fixed in base template: https://github.com/neuromation/template-base-image/blob/ca9f7cae744fb0769197c1b015b4835299583ca5/targets/Dockerfile.python36-jupyter-pytorch-tensorflow-jupyterlab#L98-L102

So we'll have difficulties if versions differ.

anayden:
Closes #117

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements.txt
+++ b/{{cookiecutter.project_slug}}/requirements.txt
@@ -1,7 +1,1 @@
-torchvision
-torch
-tqdm
-matplotlib
-numpy
-tensorboardX
-scikit-learn
+tqdm==4.32.1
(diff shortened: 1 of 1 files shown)


ayushkovskiy:
From all the packages in `requirements.txt`, we re-install all except `tqdm`  and `requests`.
- `tqdm` is a tiny package, which we can allow ourselves to install during `make setup`, let's keep it (with fixed version to follow best practices);
- `requests` -- as Lexi Bender once said, only 5% of projects need this package, let's drop it.
- all other packages are installed in `Dockerfile`, no need to re-install them again.

This all is in order to decrease time of `make setup`.

anayden:
@ayushkovskiy did I do what you mean?

mariyadavydova:
Folks, is there anything to do here? I see that there is only `tqdm` with fixed version in the requirements.txt, looks like this is what we want, isn't it? @ayushkovskiy?

PR #156, opened 2019-11-01
project_slug with dashes instead of underscores

mariyadavydova:
We use project_slug for the repo name (which is usually with dashes, not underscores) and in job names (where we can't even use underscores).


mariyadavydova:
I promise to run tests locally before submitting next time :D

PR #161, opened 2019-11-18
Fix target: `make setup-local`

ayushkovskiy:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/160


tests/e2e/test_e2e_template.py
```diff
@@ -280,7 +280,7 @@ def _test_make_run_something_useful(target: str, path: str, timeout_run: int) ->
             job_id,
             expect_patterns=["<html.*>"],
             error_patterns=["curl: .+"],
-            verbose=True,
+            verbose=False,
```
ayushkovskiy: side change: otherwise it prints a lot of trash to logs

PR #163, opened 2019-11-19
Fix test checking code availability from the notebooks

ayushkovskiy:
Due to the problems with tests https://github.com/neuromation/cookiecutter-neuro-project/issues/162, the tests didn't fail however the problem exist.

### DIFF
--- a/tests/e2e/configuration.py
+++ b/tests/e2e/configuration.py
@@ -73,7 +73,9 @@ MK_PROJECT_FILES = [PROJECT_PIP_FILE_NAME, PROJECT_APT_FILE_NAME, "setup.cfg"]
 PACKAGES_APT_CUSTOM = ["python", "expect", "figlet"]
 PACKAGES_PIP_CUSTOM = ["aiohttp==3.6", "aiohttp_security", "neuromation==19.9.10"]
 
-PROJECT_HIDDEN_FILES = {".gitkeep"}
+# TODO(artem): hidden files is a hack, see issue #93
+PROJECT_HIDDEN_FILES = {".gitkeep", "__pycache__"}
+
 PROJECT_CODE_DIR_CONTENT = {"__init__.py", "main.py"}
 PROJECT_NOTEBOOKS_DIR_CONTENT = {"Untitled.ipynb", "00_notebook_tutorial.ipynb"}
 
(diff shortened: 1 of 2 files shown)


tests/e2e/test_e2e_template.py
```diff
@@ -284,7 +286,7 @@ def _test_make_run_something_useful(target: str, path: str, timeout_run: int) ->
             job_id,
             expect_patterns=["<html.*>"],
             error_patterns=["curl: .+"],
-            verbose=True,
+            verbose=False,
```
ayushkovskiy: side change: do not overload logs with long outputs of http responses.

tests/e2e/test_e2e_template.py
```diff
@@ -133,7 +134,7 @@ def _run_make_setup_test(tmp_path: Path) -> None:
 
     expected_string = "Hello World!"
     tmp_path.mkdir(exist_ok=True)
-    out_file = (tmp_path / "out").absolute()
+    out_file = f"/tmp/out-nbconvert-{MK_PROJECT_SLUG}"
```
ayushkovskiy: this is from within the job, not on the local machine :man_facepalming:

tests/e2e/configuration.py
```diff
@@ -73,7 +73,9 @@ def unique_label() -> str:
 PACKAGES_APT_CUSTOM = ["python", "expect", "figlet"]
 PACKAGES_PIP_CUSTOM = ["aiohttp==3.6", "aiohttp_security", "neuromation==19.9.10"]
 
-PROJECT_HIDDEN_FILES = {".gitkeep"}
+# TODO(artem): hidden files is a hack, see issue #93
+PROJECT_HIDDEN_FILES = {".gitkeep", "__pycache__"}
```
ayushkovskiy: this test also fails.

tests/e2e/test_e2e_template.py
```diff
@@ -131,21 +139,21 @@ def _run_make_setup_test(tmp_path: Path) -> None:
     )
     job_id = parse_job_id(out)
 
-    expected_string = "Hello World!"
-    tmp_path.mkdir(exist_ok=True)
-    out_file = (tmp_path / "out").absolute()
```
ayushkovskiy: I was so stupid to create the file's temp directory locally and expect it to be present in the job :joy:

PR #164, opened 2019-11-19
Increase pexpect buffer for test runners

ayushkovskiy:
Getting the following errors locally:
```
   assert "Name: neuromation" in run("pip show neuromation", verbose=False)
E   assert 'Name: neuromation' in "ame: neuromation\r\nVersion: 19.9.23\r\nSummary: Neuromation Platform API client\r\nHome-page: https://neuromation.io... version 19.3.1 is available.\r\nYou should consider upgrading via the 'pip install --upgrade pip' command.\x1b[0m\r\n"
E    +  where "ame: neuromation\r\nVersion: 19.9.23\r\nSummary: Neuromation Platform API client\r\nHome-page: https://neuromation.io... version 19.3.1 is available.\r\nYou should consider upgrading via the 'pip install --upgrade pip' command.\x1b[0m\r\n" = run('pip show neuromation', verbose=False)
```
seems like the first char didn't get into the buffer.

### DIFF
--- a/tests/e2e/configuration.py
+++ b/tests/e2e/configuration.py
@@ -120,7 +120,7 @@ JOB_ID_DECLARATION_PATTERN = re.compile(
 
 # == pexpect config ==
 
-PEXPECT_BUFFER_SIZE_BYTES = 50 * 1024
+PEXPECT_BUFFER_SIZE_BYTES = 100 * 1024
 # use `sys.stdout` to echo everything to standard output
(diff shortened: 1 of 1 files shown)


ayushkovskiy:
Seems to be an issue of pexpect when tested against Python 3.6. However, I think 100 Kb of buffer is more stable, Plus, since we discovered this issue, let's fix it (Python 3.6 is probable to be tested on).

PR #184, opened 2019-11-27
Fix CMD for user's base image

ayushkovskiy:
Great catch by @atselousov!

Sorry @mariyadavydova for asking you one day to change the command for `make setup` job from `sleep infinity` to `sleep 1h` -- it appeared to be that `docker commit` (which is used in `neuro save`, which is used by `make setup`) sets the `CMD` passed to the container (in this case, `sleep 1h`) to the new image. Then, if one starts a job from this newly created image, and does not specify the command (`neuro run image:my-base-image`), the command will be `sleep 1h`!


mariyadavydova:
I think that we didn't expect anyone run the base image without the command. Thank you for fixing it :)

PR #191, opened 2019-11-29
[tests] Fix exit code check

ayushkovskiy:
Fix of exit-code check introduced in https://github.com/neuromation/cookiecutter-neuro-project/pull/175. Will close https://github.com/neuromation/cookiecutter-neuro-project/issues/162

### DIFF
--- a/tests/e2e/helpers/runners.py
+++ b/tests/e2e/helpers/runners.py
@@ -121,7 +121,7 @@ def _run_once(
     to log (also to dump all child process' output to the handler defined
     in `PEXPECT_DEBUG_OUTPUT_LOGFILE`).
     By default the method throws ExitCodeException if the process is killed or
-    exits with non-zero exit code. Passing allow_nonzero_exitcode=True suppresses this
+    exits with non-zero exit code. Passing assert_exit_code=False suppresses this
     behavior.
     >>> # Expect the first and the last output:
     >>> _run_once("echo 1 2 3", expect_patterns=[r'1 \d+', '3'], verbose=False)
--- a/tests/e2e/helpers/runners.py
+++ b/tests/e2e/helpers/runners.py
@@ -199,14 +199,16 @@ def _run_once(
                 output += chunk
         if assert_exit_code:
             if child.isalive():
-                # Short sleep to allow make command to complete
-                log_msg(f"Sleeping for {cmd}", logger=LOGGER.info)
-                time.sleep(1)
+                # flush process buffer
+                child.read()
+                # wait for child to exit
+                log_msg(f"Waiting for {cmd}", logger=LOGGER.info)
+                child.wait()
             child.close(force=True)
             if child.status:
                 need_dump = True
                 if child.signalstatus is not None:
-                    log_msg(f"{cmd} was killed", logger=LOGGER.warning)
+                    log_msg(f"{cmd} was killed via signal", logger=LOGGER.warning)
                 raise ExitCodeException(child.status)
     finally:
         if detect_new_jobs:
--- a/tests/e2e/helpers/runners.py
+++ b/tests/e2e/helpers/runners.py
@@ -40,7 +40,7 @@ def run(
     error_patterns: t.Sequence[str] = DEFAULT_ERROR_PATTERNS,
     verbose: bool = True,
     detect_new_jobs: bool = True,
-    allow_nonzero_exitcode: bool = False,
+    assert_exit_code: bool = True,
 ) -> str:
     """
     This procedure wraps method `_run`. If an exception raised, it repeats to run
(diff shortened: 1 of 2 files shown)


ayushkovskiy:
@anayden valid test on `make setup` fails with exit code 1: https://circleci.com/gh/neuromation/cookiecutter-neuro-project/855. Most likely, the problem is with pexpect's API we use.

Please take finish PR as it's intended to fix bug introduced in https://github.com/neuromation/cookiecutter-neuro-project/pull/175.

tests/e2e/helpers/runners.py
```diff
@@ -197,7 +197,7 @@ def _run_once(
             finally:
                 chunk = _get_chunk(child)
                 output += chunk
-        if allow_nonzero_exitcode:
+        if not allow_nonzero_exitcode:
```
ayushkovskiy: I'd also rename this parameter to more clear `assert_exit_code`

tests/e2e/helpers/runners.py
```diff
@@ -56,7 +56,7 @@ def run(
                 verbose=verbose,
                 detect_new_jobs=detect_new_jobs,
                 timeout_s=timeout_s,
-                allow_nonzero_exitcode=False,
+                allow_nonzero_exitcode=allow_nonzero_exitcode,
```
ayushkovskiy: another bug ^

ayushkovskiy:
needs rebase on the newest master so that stage `cleanup e2e` does not hang

tests/e2e/helpers/runners.py
```diff
@@ -197,8 +197,12 @@ def _run_once(
             finally:
                 chunk = _get_chunk(child)
                 output += chunk
-        if allow_nonzero_exitcode:
-            child.close()
+        if assert_exit_code:
+            if child.isalive():
+                # Short sleep to allow make command to complete
+                log_msg(f"Sleeping for {cmd}", logger=LOGGER.info)
+                time.sleep(1)
```
ayushkovskiy: what's the reason for this code? please provide either explanation/proof that we need it or let's drop it.

ayushkovskiy: and also explanation of this exact sleep delay.

anayden: I thought the comment above these lines was clear, my bad.

My experiments indicate that after we find all required matching patterns it takes a tiny bit of time for `make` process to complete. If we call `close()` immediately, the process gets killed and hence returns a non-zero `signalstatus`.

As far as _exact sleep delay is concerned_, I'm afraid there's no such thing in asynchronous world. There's only good enough and not good enough delays. This one proved to be just fine in all my tests.

ayushkovskiy: what I know is that any sleep is a potential source of flakiness, this is my concern. As an alternative, can we use `child.close(force=False)`? Will it wait until the child is closed?

anayden: > any sleep is a potential source of flakiness

I'm very well aware of that and agree with you. `close(force=False)` still sends `SIGTERM` to process which makes `signalstatus` non-zero, I tested that as well.

ayushkovskiy: should we use [`wait`](https://pexpect.readthedocs.io/en/stable/api/pexpect.html#pexpect.spawn.wait) instead?

anayden: Already updated. Now dealing with NFS flakiness in an unrelated test

anayden: All tests pass, yay!

tests/e2e/helpers/runners.py
```diff
@@ -197,12 +197,18 @@ def _run_once(
             finally:
                 chunk = _get_chunk(child)
                 output += chunk
-        if allow_nonzero_exitcode:
-            child.close()
+        if assert_exit_code:
+            if child.isalive():
+                # flush process buffer
+                child.read()
```
ayushkovskiy: let's save this last string: `output += child.read()`

tests/e2e/helpers/runners.py
```diff
@@ -197,12 +197,18 @@ def _run_once(
...
+                # wait for child to exit
+                log_msg(f"Waiting for {cmd}", logger=LOGGER.info)
+                child.wait()
+            child.close(force=True)
```
ayushkovskiy: thank you, it seems to work! the last thing: let's add a few doctests here covering following situation:
1. `cmd="echo 1; sleep 2; echo 2", expected_patterns=["1"]`, so we check that the job's exit code is `0` and its output contains `2`
2. same `cmd`, `expected_patterns=[]`
3. `cmd="echo 1"` -- we check the branch where `child.isalive()` is False

ayushkovskiy: @anayden can we have these changes tested?

ayushkovskiy: We need this functionality. So let's merge now and add tests later: https://github.com/neuromation/cookiecutter-neuro-project/issues/230

tests/e2e/helpers/runners.py
ayushkovskiy: fixed.

ayushkovskiy:
tests fixed and actualized with changes in master.

PR #202, opened 2019-12-03
[Makefile] Add target `make format`

ayushkovskiy:
very useful, we have it in almost all our python projects.


AlekseySh:
I hate blake

mariyadavydova:
@AlekseySh What's wrong with it?

ayushkovskiy:
I think in Template we should follow best practices of programming, and black+lint are best practices. Since we have lint, I think we should have black as well.

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -196,10 +196,17 @@ kill: kill-training kill-jupyter kill-tensorboard kill-filebrowser
 setup-local:  ### Install pip requirements locally
 	$(PIP) -r requirements.txt
 
+.PHONY: format
+format:  ### Run static code analysis locally
```
mariyadavydova: Incorrect help message. I assume this is "Automatically format the code" rather than run the analysis.

PR #204, opened 2019-12-03
[Makefile] Run `upload-code` before `training`

ayushkovskiy:
as requested in https://docs.google.com/document/d/13c_eYG2v4wNnp9RIhE_gXCSTvEw_i_HQzK7djEsSOjw/

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -117,7 +117,7 @@ clean: clean-code clean-data clean-notebooks
 ##### JOBS #####
 
 .PHONY: training
-training:  ### Run a training job
+training: upload-code  ### Run a training job
 	$(NEURO) run \
 		--name $(TRAINING_JOB) \
 		--preset $(TRAINING_MACHINE_TYPE) \


ayushkovskiy:
However I would still vote for removing dependencies between targets: https://github.com/neuromation/cookiecutter-neuro-project/issues/118

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -117,7 +117,7 @@ clean: clean-code clean-data clean-notebooks
 ##### JOBS #####
 
 .PHONY: training
-training:  ### Run a training job
+training: upload-code  ### Run a training job
```
atselousov: Is it a good idea to merge two different commands into one? What to do if code was modified on storage before training?

We have ["discussion"](https://docs.google.com/document/d/13c_eYG2v4wNnp9RIhE_gXCSTvEw_i_HQzK7djEsSOjw/edit?disco=AAAADy6MyHc) about it.

mariyadavydova: Indeed. There are several different use-cases, and we discussed them today with 2 of ML engineers. 

One of them opts for adding this dependency because he forgets to re-upload the code, another one opts against this dependency because he likes to modify the code on the server directly. The first one agreed that drop this request for now.

ayushkovskiy: I think @atselousov's concern is resolved since indeed `neuro cp --update` does not re-write a file if it was updated later:
```
[ay@archlinux ~]$ cat test.txt
123
[ay@archlinux ~]$ n run -s cpu-small -v storage::/mnt:rw ubuntu 'bash -c "echo 456 > /mnt/test.txt && cat /mnt/test.txt"'
...             
456
[ay@archlinux ~]$ n cp --update test.txt storage:
Copy 'file:///home/ay/test.txt' => 'storage://artemyushkovskiy/test.txt'
[ay@archlinux ~]$ n run -s cpu-small -v storage::/mnt:rw ubuntu "cat /mnt/test.txt"
...        
456
[ay@archlinux ~]$ 
```

@mariyadavydova I vote for consistency: since we already have other inter-target dependencies, we obviously lack dependency `make training` on `make upload-code`. Once we decide to [remove inter-target dependencies](https://github.com/neuromation/cookiecutter-neuro-project/issues/118), we will remove this dependency as well.

So let's merge this.

mariyadavydova: OK. This is long discussion, I hope, we come to some conclusion one day. Let's merge.

PR #210, opened 2019-12-03
[Makefile][Readme] Add argument to `make training`

ayushkovskiy:
Example:
```
$ make train  # runs job training-project-name-base
$ make kill-train   # kills job training-project-name-base
$ make train  RUN=second   # runs job training-project-name-second
$ make kill-train  RUN=second  # kills job training-project-name-second
$ export RUN=3
$ make train   # runs job training-project-name-3
$ make kill-train   # kills job training-project-name-3
$ make kill-train-all
```
All training job names are dumped to `./.train_jobs` and this file is used by `make kill-train-all`


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -116,10 +116,12 @@ clean: clean-code clean-data clean-notebooks
 
 ##### JOBS #####
 
+EXPERIMENT?=0
```
atselousov: Wouldn't it be more useful to name the default train project for example `base`, i.e. set `EXPERIMENT=base`? Meaningful names help to figure out what each job does if you start a lot of them. For me, it's not clear that `$(TRAINING_JOB)-0` means.

Maybe, it has sense to add this postfix to other jobs which require `gpu` resources (e.g. `jupyter`).

It also would be a good idea to describe this functional and how to modify your own commands in `README` file of an inited project.

ayushkovskiy: Sure, thank you!

Also, thanks for pointing out on the documentation, I totally forgot about it...

ayushkovskiy: cool idea, thanks!

ayushkovskiy: fixed

{{cookiecutter.project_slug}}/README.md
```diff
@@ -91,11 +91,34 @@ When jobs with HTTP interface are executed (for example, with Jupyter Notebooks
...
+TRAINING_COMMAND="python /path/to/script.py"
+```
+And then, just run `make training`
+
+
+You can also run multiple training experiments simultaneously. For that, set up `EXPERIMENT` environment variable:
+```
+make training  EXPERIMENT=check-new-idea
+```
+Please note that this label becomes a postfix of the job name, which may contain only alphanumeric characters
```
atselousov: Please, add information on how to kill a job with postfix in a name. It's unclear now that you need to use the same postfix to kill a training job.

ayushkovskiy: right! thank you!

ayushkovskiy: fixed

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -116,10 +116,13 @@ clean: clean-code clean-data clean-notebooks
...
 ##### JOBS #####
 
+EXPERIMENT?=base
+
 .PHONY: training
 training:  ### Run a training job
 	$(NEURO) run \
-		--name $(TRAINING_JOB) \
+		--name $(TRAINING_JOB)-$(EXPERIMENT) \
+		--description "$(PROJECT_POSTFIX): training job" \
```
atselousov: Is it possible when different projects have the same job description?

ayushkovskiy:
*Resumé*

Problem:
- we need to be able to run multiple concurrent trains
- no way to pass make-target arguments except env var
- even if syntax `make training EXPERIMENT=second` is OK, we need to be able to kill all training jobs submitted

Possible solutions:
1. keep the same syntax `make kill-training EXPERIMENT=second`, but also keep track of all training jobs run to a file, and then during `make kill-all` read the job list from that file and kill jobs.
2. Use `--description` field as proposed in current version of this PR. The problem is that it's not enough to `grep` by project-id, to be sure not to kill other project's jobs, we need a randomized label. So let's add a random label to the description of *all jobs* run in current project? In this approach, we can transform `make kill-all` to `neuro kill $(neuro -q ps --description="${PROJECT_NAME} label:${RANDOM_LABEL}`).

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -46,18 +46,22 @@ PRESET?=gpu-small
 HTTP_AUTH?=--http-auth
 
 # Command to run training inside the environment. Example:
-TRAIN_CMD=python -u $(CODE_DIR)/train.py --data $(DATA_DIR)
+TRAIN_CMD?=python -u $(CODE_DIR)/train.py --data $(DATA_DIR)
```
ayushkovskiy: Important bug fix: without this, `export TRAIN_CMD` (as stated in Readme) does not work

{{cookiecutter.project_slug}}/README.md
```diff
@@ -216,8 +216,35 @@ the platform, you may disable the authentication updating this line to `HTTP_AUT
...
+
+Note, in this case we use single quotes so that local `bash` does not resolve environment variables. 
+You can assume that training command `TRAIN_CMD` runs in the project's root directory.
+
+
+### Multiple training jobs
+
+You can run multiple training experiments simultaneously by setting up `RUN` environment variable:
+```
+make training RUN=new-idea
```
ayushkovskiy: outdated here and onwards: should be `make train`

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -268,10 +271,14 @@ port-forward-develop:  ### Forward SSH port to localhost for remote debugging
 kill-develop:  ### Terminate the development job
 	$(NEURO) kill $(DEVELOP_JOB)
 
+RUN?=base
+
 .PHONY: train
-train: _check_setup upload-code upload-config   ### Run a training job
+train: _check_setup upload-code upload-config   ### Run a training job (set up env var 'RUN' to specify the training job),
+	@[ -f $(_TRAIN_JOB_FILE) ] && grep -qxF '$(TRAIN_JOB)-$(RUN)' $(_TRAIN_JOB_FILE) || \
+		{ echo '$(TRAIN_JOB)-$(RUN)' >> $(_TRAIN_JOB_FILE); echo "Job name '$(TRAIN_JOB)-$(RUN)' saved to file '$(_TRAIN_JOB_FILE)'"; }
```
ayushkovskiy: better: `echo '$(TRAIN_JOB)-$(RUN)' >> $(_TRAIN_JOB_FILE) && echo "Job name ...`

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -285,12 +292,20 @@ train: _check_setup upload-code upload-config   ### Run a training job
...
 .PHONY: kill-train
-kill-train:  ### Terminate the training job
-	$(NEURO) kill $(TRAIN_JOB)
+kill-train:  ### Terminate the training job (set up env var 'RUN' to specify the training job)
+	$(NEURO) kill $(TRAIN_JOB)-$(RUN)
+
+.PHONY: kill-train-all
+kill-train-all:  ### Terminate all training jobs you have submitted (set up env var 'RUN' to specify the training job)
+ifneq ($(wildcard $(_TRAIN_JOB_FILE)),)
+	$(NEURO) kill $$(cat $(_TRAIN_JOB_FILE)) && rm $(_TRAIN_JOB_FILE)
```
ayushkovskiy: Note: currently, `neuro kill` always returns zero exit code, so the file will always be removed

ayushkovskiy:
All your comments considered, thank you! 

Plus, I changed the logic: the file storing train-job-ids is now deleted only if no errors was during `neuro kill`. This is needed not to lose jobs running (otherwise `make kill-all` might forget about them).

Plus, added a test on invalid job name postfix: `make train RUN=...`

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -285,12 +292,20 @@ train: _check_setup upload-code upload-config   ### Run a training job
...
 .PHONY: kill-train
-kill-train:  ### Terminate the training job
-	$(NEURO) kill $(TRAIN_JOB)
+kill-train:  ### Terminate the training job (set up env var 'RUN' to specify the training job)
+	$(NEURO) kill $(TRAIN_JOB)-$(RUN)
+
+.PHONY: kill-train-all
+kill-train-all:  ### Terminate all training jobs you have submitted (set up env var 'RUN' to specify the training job)
+ifneq ($(wildcard $(_TRAIN_JOB_FILE)),)
+	$(NEURO) kill $$(cat $(_TRAIN_JOB_FILE)) 2>&1 | grep -Evc "(ERROR: |Cannot kill job)" && rm $(_TRAIN_JOB_FILE)
```
ayushkovskiy: unfortunately, `neuro kill` exits with `0` even if it failed to kill a job :(

ayushkovskiy: https://github.com/neuromation/platform-client-python/issues/1272

PR #214, opened 2019-12-05
[Makefile] Rename `make ps` -> `make list-jobs`

ayushkovskiy:
for a regular user, `ps` means nothing. Moreover, `make ps` makes no sense if it runs only `neuro ps`. When renamed, it makes more sense.

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -203,6 +203,6 @@ lint:  ### Run static code analysis locally
 
 ##### MISC #####
 
-.PHONY: ps
-ps:  ### List all running and pending jobs
+.PHONY: list-jobs
+list-jobs  ### List all running and pending jobs
 	$(NEURO) ps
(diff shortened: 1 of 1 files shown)


AlekseySh:
probably we have to grep only jobs related to this project?
and better name is `make list-project-jobs`

ayushkovskiy:
> and better name is make list-project-jobs

I think this name is too long, since everything in Template relates to the project, let's drop this word.

ayushkovskiy:
@mariyadavydova @AlekseySh 

Added grep, please re-review

ayushkovskiy:
note, `grep` may be not available on Windows

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -203,6 +203,6 @@ lint:  ### Run static code analysis locally
 
 ##### MISC #####
 
-.PHONY: ps
-ps:  ### List all running and pending jobs
-	$(NEURO) ps
+.PHONY: list-jobs
+list-jobs:  ### List all running and pending jobs submitted in current project
+	$(NEURO) ps | grep $(PROJECT_POSTFIX) || true
```
ayushkovskiy: What if there's another job with the same name but from another project?..

ayushkovskiy:
Closing as `| grep "-$(PROJECT_POSTFIX)"` might include a non-project job if its `neuro-ps` output contains `-$(PROJECT_POSTFIX)`. Let's keep it simple for now.

PR #237, opened 2019-12-07
[Makefile][Template] Add config directory

ayushkovskiy:
a separate mount is needed to store secrets there


ayushkovskiy:
thank you @mariyadavydova , your comments fixed!

Issue #211, opened 2019-12-04
Integrate with google cloud storage


ayushkovskiy:
see https://github.com/neuromation/platform-web/pull/119

ayushkovskiy:
done

PR #243, opened 2019-12-10
[Makefile] Google Cloud SDK integration: full automation

ayushkovskiy:
Implements instructions from https://github.com/neuromation/platform-web/pull/119

The main idea of `make gcloud-setup` is not to perform any change via `gcloud` or `gsutil`, but only to check the configuration automatically and to suggest user command to fix. Once it succeeds, the secret file and env var configuration files are stored to the `./config/` directory. Then, once the job started, if the secret file exists, `gcloud auth` will be called.

Idea for future: perform encryption of secret files stored in `./config` so that shared jobs don't reveal secret tokens.

TODO: Some tests will be implemented once the Makefile is reviewed.

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -54,7 +54,7 @@ LOCAL_PORT?=2211
 GCP_SECRET_FILE?=neuro-job-key.json
 GCP_PROJECT_ID?=
 GCP_SERVICE_ACCOUNT_NAME?=
-GCP_BUCKET_NAME?=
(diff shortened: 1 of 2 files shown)


astaff:
Thank you. Let's keep it open for now as this seems to be out of scope for the current release. It was not clear from the description of #211 what it look like from the user's standpoint and why are we choosing to  wrap gsutil in Maketargets and what other options have been considered. We will get back to this in the next release as an improvement to the documentation based on the feedback.

ayushkovskiy:
Will get back to it later.

PR #287, opened 2019-12-20
[Template] Add dependencies mypy+isort+black+flake

ayushkovskiy:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/283

Before, `tqdm` in `requirements.txt` was just an example. 
1. We'll install it to the base image
2. No tests rely on it.

Question: Should we pre-install these dependencies also to the base image in order to shorten `make setup` time?


mariyadavydova:
Let's keep them here. Otherwise, an empty file is a bad example.

PR #289, opened 2019-12-20
[Makefile] Stream tqdm logs for all training jobs via `neuro exec`

ayushkovskiy:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/288

### DIFF
--- a/tests/e2e/test_e2e_template.py
+++ b/tests/e2e/test_e2e_template.py
@@ -503,10 +503,7 @@ def test_make_download_all() -> None:
 def test_make_train_defaults(env_neuro_run_timeout: int) -> None:
     _run_make_train(
         env_neuro_run_timeout,
-        expect_patterns=[
-            _get_pattern_status_succeeded_or_running(),
-            "Your training script here",
-        ],
+        expect_patterns=[_get_pattern_status_succeeded_or_running()],
     )
 
 
@@ -516,7 +513,7 @@ def test_make_train_custom_command(
 ) -> None:
     cmd = env_py_command_check_gpu
     cmd = cmd.replace('"', r"\"")
-    cmd = f"'python -W ignore -c \"{cmd}\"'"
+    cmd = f"bash -c 'sleep 5 && python -W ignore -c \"{cmd}\"'"
     monkeypatch.setenv("TRAINING_COMMAND", cmd)
     # NOTE: tensorflow outputs a lot of debug info even with `python -W ignore`.
     #  To disable this, export env var `TF_CPP_MIN_LOG_LEVEL=3`
@@ -584,7 +581,7 @@ def test_make_train_invalid_name(
     cmd_prtn = "make train TRAIN_CMD='sleep 1h' RUN={run}"
 
     with finalize(f"neuro kill {job_valid}"):
-        cmd_valid = cmd_prtn.format(run=exp_valid)
+        cmd_valid = cmd_pattern.format(run=exp_valid)
         with measure_time(cmd_valid, TIMEOUT_NEURO_RUN_CPU):
             run(
                 cmd_valid,
@@ -592,7 +589,7 @@ def test_make_train_invalid_name(
                 assert_exit_code=False,
             )
 
-        cmd_invalid = cmd_prtn.format(run=exp_invalid)
+        cmd_invalid = cmd_pattern.format(run=exp_invalid)
         with measure_time(cmd_invalid, TIMEOUT_NEURO_RUN_CPU):
             run(
                 cmd_invalid,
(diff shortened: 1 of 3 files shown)


ayushkovskiy:
If a separate target `logs-tqdm-train` does not go, we could optionally print tqdm output for each train job if an env var `PRINT_TQDM` is set

{{cookiecutter.project_slug}}/requirements.txt
```diff
@@ -1 +0,0 @@
-tqdm==4.32.1
```
ayushkovskiy: it's installed in base image already

mariyadavydova:
I would prefer to see the output in `make train` directly. The problem is that people almost never read the docs and thus will ignore the additional option even if we describe it accurately.

Therefore, may we show `stdout` and `stderr` from a job simultaneously?

ayushkovskiy:
`cat` steals streams from `neuro log`.

`2>&1 | tee /tmp/output.txt`

tests/e2e/test_e2e_template.py
```diff
@@ -478,6 +478,40 @@ def _run_make_train_test(neuro_run_timeout: int, expect_patterns: List[str]) ->
...
 
+@pytest.mark.run(order=STEP_RUN)
+def test_make_train_tqdm(env_var_preset_cpu_small: str) -> None:
+    _test_make_train_tqdm()
+
+
+@try_except_finally(f"neuro kill {MK_TRAIN_JOB}")
+def _test_make_train_tqdm() -> None:
+    py_cmd = "import time, tqdm; [time.sleep(0.01) for _ in tqdm.tqdm(range(10000))]"
+    bash_cmd = f"python -W ignore -c '{py_cmd}'"
```
ayushkovskiy: wrong `-W ignore` here

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -85,7 +85,7 @@ NEURO?=neuro
 
 
 ifeq (${TRAIN_STREAM_LOGS}, yes)
-	_TRAIN_WAIT_START_OPTION=--wait-start
+	_TRAIN_WAIT_START_OPTION=--wait-start --detach
```
mariyadavydova: Given this change we do not need to keep detach option here. You always want to detach your train job. 

Also, why is there an underscore before the name? Looks like we do not use such a naming scheme anywhere in Makefile (e.g. we use OPTION_GCP_CREDENTIALS). Could you please rename it to TRAIN_WAIT_START?

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -314,7 +317,7 @@ kill-train:  ### Terminate the training job (set up env var 'RUN' to specify the
 kill-train-all:  ### Terminate all training jobs you have submitted (set up env var 'RUN' to specify the training job)
 ifneq ($(wildcard $(_TRAIN_JOB_FILE)),)
 	$(NEURO) kill $$(cat $(_TRAIN_JOB_FILE) | uniq -u)
-	@echo -e "Please check that jobs listed in '$(_TRAIN_JOB_FILE)' are actually deleted.\nUse 'neuro ps' to verify."
+	@echo -e "Please check that jobs listed in '$(_TRAIN_JOB_FILE)' are actually killed and delete this file manually.\nUse 'neuro ps' to verify."
```
mariyadavydova: Killed -> terminated

{{cookiecutter.project_slug}}/Makefile
ayushkovskiy: thanks!

{{cookiecutter.project_slug}}/Makefile
ayushkovskiy: thanks

Makefile
```diff
@@ -30,7 +30,7 @@ test_unit:
 
 .PHONY: test_e2e_dev
 test_e2e_dev:
-	PRESET=cpu-small NEURO=$(NEURO_COMMAND)  pytest -s -v --environment=dev --tb=short tests/e2e
+	PRESET=cpu-small NEURO=$(NEURO_COMMAND)  pytest -s -v --environment=dev --tb=short tests/e2e -k "make_setup or tqdm"
```
ayushkovskiy: DEBUGGING

tests/e2e/test_e2e_template.py
```diff
@@ -516,7 +513,7 @@ def test_make_train_custom_command(
 ) -> None:
     cmd = env_py_command_check_gpu
     cmd = cmd.replace('"', r"\"")
-    cmd = f"'python -W ignore -c \"{cmd}\"'"
+    cmd = f"bash -c 'sleep 5 && python -W ignore -c \"{cmd}\"'"
```
ayushkovskiy: job needs to be running to use `neuro exec`

PR #291, opened 2019-12-20
[tests] Integrate instafail

ayushkovskiy:
Integrate [pytest-instafail](https://github.com/pytest-dev/pytest-instafail): show failures and errors instantly instead of waiting until the end of test session


ayushkovskiy:
doesn't work as expected :(

PR #296, opened 2019-12-23
[Makefile] [maybe] Allow to pass any option to neuro-run commands

ayushkovskiy:
Not sure if it's a good idea.

Motivation: 
1. For [hyper-parameter search](https://github.com/neuromation/ml-recipe-hyper-search/compare/dev), we need to pass `--no-wait-start`. 
2. Also, we already have `HTTP_AUTH`. 

So maybe it's better to merge all these extra options to a single variable, `NEURO_RUN_EXTRA_OPTIONS`? I'm in doubts.

TODO: if this idea is accepted, we need to get rid of `HTTP_AUTH` in this very PR.

### DIFF
--- a/tests/e2e/test_e2e_template.py
+++ b/tests/e2e/test_e2e_template.py
@@ -443,6 +443,18 @@ def test_make_download_results() -> None:
 
 
 @pytest.mark.run(order=STEP_RUN)
+def test_make_train_extra_option_works(env_neuro_run_timeout: int) -> None:
+    run(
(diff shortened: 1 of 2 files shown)


mariyadavydova:
I don't see a problem we are solving here. `HTTP_AUTH` speaks for itself. `NEURO_RUN_EXTRA_OPTIONS` makes reading Makefile even harder. I wouldn't change anything here for now.

ayushkovskiy:
ok

PR #297, opened 2019-12-23
[Makefile] Fix `make setup` UX issue

ayushkovskiy:
1. Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/242
2. Runs `neuro mkdir` with parameter `--parent` during `make setup`

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -97,7 +97,12 @@ help:
 
 .PHONY: setup
 setup: ### Setup remote environment
-	$(NEURO) kill $(SETUP_JOB) >/dev/null 2>&1
+	$(NEURO) mkdir -p $(PROJECT_PATH_STORAGE) \
+		$(PROJECT_PATH_STORAGE)/$(CODE_DIR) \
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -97,7 +97,12 @@ help:
 
 .PHONY: setup
 setup: ### Setup remote environment
-	$(NEURO) kill $(SETUP_JOB) >/dev/null 2>&1
+	$(NEURO) mkdir --parents $(PROJECT_PATH_STORAGE) \
```
mariyadavydova: Oh yeah! I knew, there should have been an option for that 😀

Actually, if there is such an option, let's better update the `upload-*` and `clean-*` targets:
* Create a directory with `--parents` on each upload.
* Remove a directory on each clean.
* Create only a project root directory during the setup.

ayushkovskiy: > Remove a directory on each clean

I disagree: recently, we changed `clean-*` targets not to remove the whole directory, but only the content, in order to preserve project's structure. So let's discuss it.

PR #307, opened 2019-12-26
[Makefile] Hyper-parameter search via W&B

ayushkovskiy:
Based on https://github.com/neuromation/ml-recipe-hyper-search/pull/1


ayushkovskiy:
Note: in tests, we need to check that training script has started (that `train.py` has run and printed the default value). Without it, I don't trust tests.

tests/e2e/test_e2e_template.py
```diff
@@ -526,12 +527,13 @@ def test_make_train_multiple_experiments(
...
-            assert_exit_code=False,
-            detect_new_jobs=False,
-        )
-        assert not jobs_left
+
+        # TODO: neuro-ps + grep is blocked by neuro issue:
+        #  https://github.com/neuromation/platform-client-python/issues/1189
+        # hyper_ls = f"bash -c 'neuro ps | grep {mk_train_job('-hyper-')}'"
+        # running = run(hyper_ls, assert_exit_code=False).splitlines()
+        # assert not running, f"expect no hyper-jobs: {running}"
```
ayushkovskiy: this change and onwards -- currently not failing but potentially incorrect until we fix `neuro`

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -302,13 +305,15 @@ port-forward-develop:  ### Forward SSH port to localhost for remote debugging
 kill-develop:  ### Terminate the development job
 	$(NEURO) kill $(DEVELOP_JOB)
 
+
 RUN?=base
+_SWEEP_DESCRIPTION_POSTFIX?=
```
ayushkovskiy: internal variable, not to be documented or used by the user => starts with `_`

tests/e2e/test_e2e_template.py
```diff
@@ -630,6 +632,56 @@ def test_make_train_tqdm(env_var_preset_cpu_small: str) -> None:
         run("make kill-train", detect_new_jobs=False)
 
 
+@pytest.mark.run(order=STEP_RUN)
+def test_make_hypertrain(
+    monkeypatch: Any,
```
anayden: `monkeypatch` is not used, is it?

ayushkovskiy: ok

ayushkovskiy:
for grep version 2.5, option `-P` is unavailable
```
grep (BSD grep) 2.5.1-FreeBSD

alexeyshab@LexBook:~/PycharmProjects/check-sweep|
⇒  make hypertrain 
Creating W&B Sweep...
Using variable: WANDB_SWEEP_CONFIG_FILE='wandb-sweep.yaml'
Using W&B sweep file: ./src/wandb-sweep.yaml
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
Creating sweep from: src/wandb-sweep.yaml
make: *** [hypertrain] Error 2
```
thanks @AlekseySh for finding this.

anayden:
It's not `for grep version 2.5`, it's rather GNU grep (linux) vs BSD grep (macOS) difference. I see issues like that on the mac all the time.

mariyadavydova:
I have a meta question about W&B hyper-parameter tuning: here we run 3 jobs simultaneously, while the example implies  running total of 12 jobs (and normally the process requires running a lot of jobs with different parameters). Which part of this integration is responsible for running new jobs as soon as the old ones are over?

ayushkovskiy:
THis PR uses a lot of linux utilities, it needs to be tested on MacOS X carefully

tests/e2e/samples/config/wandb-token.txt.enc
```diff
@@ -0,0 +1 @@
+gAAAAABeBB_ozBhesCejG0-4LhL7h6vrf5bLnCCW_Dzgt8YNj3hcujUFiVHUVxFyCArsghIy_zVdeCfDQ7P3uG8kS4bmyMBCnfb3XdAVxXFE9eJr5xAEGb651W30JPkYoIStZON4_d_V
```
ayushkovskiy: my personal token encrypted

tests/e2e/test_e2e_template.py
```diff
@@ -662,6 +664,51 @@ def test_make_train_tqdm(env_var_preset_cpu_small: str) -> None:
...
+        jobs = parse_jobs_ids(out, expect_num=n)
+
+        with finalize(*(f"neuro kill {job}" for job in jobs)):
+            for job in jobs:
+                run(
+                    f"neuro logs {job}",
+                    expect_patterns=[
+                        r"Successfully logged in to Weights & Biases!",
+                        r"wandb: Starting wandb agent",
+                        r"Running runs:",
```
ayushkovskiy: Yes, this W&B says :D

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -331,6 +335,60 @@ endif
...
+	$(NEURO) cp ./wandb/settings $(PROJECT_PATH_STORAGE)/wandb/
+	@echo "Running ${N_HYPERPARAM_JOBS} jobs of sweep ${SWEEP_REAL}..."
+	for i in $$(seq 1 ${N_HYPERPARAM_JOBS}) ; do \
+		echo -e "\nStarting job $${i}..." ; \
+		make train \
+			TRAIN_CMD="wandb agent ${SWEEP_REAL}" \
+			RUN=${SWEEP_REAL}-$${i} \
+			RESULTS_DIR_STORAGE=$(RESULTS_DIR_STORAGE)/sweep-${SWEEP_REAL} \
+			_SWEEP_DESCRIPTION_POSTFIX=":${SWEEP_REAL}" \
+			_EXTRA_VOLUME_OPTION="--volume $(PROJECT_PATH_STORAGE)/wandb:$(PROJECT_PATH_ENV)/wandb:rw" \
```
ayushkovskiy: If we don't put the file `./wandb/settings` created after `wandb sweep ...`, wandb won't use the correct project (and possibly other settings stored in this file), so the command `wandb agent <swipe>` won't work from within the job.

ayushkovskiy:
Discovered that in order for the agents to work fine, we need to pass the project's wandb settings (in `<project>/wandb/`) -- via mount.

ayushkovskiy:
# NOTE: We test hypertrain only for 1 job because otherwise the test becomes
    # too flaky, for example: the first job runs successfully (for current test
    # it takes around 25 seconds to do full hyper-parameter search by 1 job).
    # Meanwhile, the second job can be easily pending for longer than 30 seconds,
    # so at the point when it gets running, the sweep is already not running.
    # E.g., see run https://circleci.com/gh/neuromation/cookiecutter-neuro-project/3141

ayushkovskiy:
This PR also contains fixes that stabilize tests.

PR #311, opened 2019-12-26
[Makefile] Upload/Download/Clean all subdirs equally

mariyadavydova:


mariyadavydova:
Thank you! Agree, let's merge it later.

PR #316, opened 2019-12-29
[tests] Give flaky tests on `neuro exec` 2 attempts

ayushkovskiy:
temporarily patches https://github.com/neuromation/cookiecutter-neuro-project/issues/280


ayushkovskiy:
on dev all tests fail with timeout. will merge to test on staging

PR #334, opened 2020-01-29
[tests] Stabilize tests

ayushkovskiy:
Important changes **in bold**.
1. **Add missing doctests on test runners (they were not run before)**
2. Fix a mistake in test runners (error patterns were not asserted)
3. Run unit test in a temp directory instead of `..` (closes https://github.com/neuromation/cookiecutter-neuro-project/issues/126)
4. Install less custom pip and apt packages (first, there were pip conflicts in versions of neuromation and aiohttp, second, there's no need to install many packages)
5. Fix import fernet -> cryptography (somehow fernet stopped working)
6. Improve logging: print env vars because they matter
7. Add test functionality to skip default error pattern search (for tests that test failure scenarios)
8. **Add test functionality to re-run only if a certain error substring was found** (for now, it's only `"stale NFS file handle"` on every target that runs a job).
9. **Temporarily ignore all timeouts (see https://github.com/neuromation/cookiecutter-neuro-project/issues/333)**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,7 @@ jobs:
           name: run unit tests
           command: |
             . venv/bin/activate
+            make test_doctest
             make test_unit
       - run:
           no_output_timeout: 20m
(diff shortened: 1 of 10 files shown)


ayushkovskiy:
Tested with 7 concurrent test builds against `dev`

PR #336, opened 2020-01-30
[tests] Cleanup tests

ayushkovskiy:
1. With a hack, fix conditional repeater (function `run()`)
2. Increase pexpect buffer size up to 10 Mb
3. minor changes


ayushkovskiy:
(1) is not fixed. Will re-work it properly later.

PR #342, opened 2020-01-31
[Makefile] Prepare for Windows: rework 'make hypertrain'

ayushkovskiy:
1. in `make setup`, get rid of the loop over `$(PROJECT_FILES)` (just more readable, nothing about Windows)
2. Get rid of the extraneous var `RESULTS_DIR_STORAGE?=$(PROJECT_PATH_STORAGE)/$(RESULTS_DIR)`: In Windows, we haven't found a way to run `make <target>` from another make target. Thus, we won't use `make train` from `make hypertrain`, so there's no need to overload var `RESULTS_DIR_STORAGE`.
3. Split `make hypertrain` to two targets: `make pre-hypertrain` + `make hypertrain`. This is done in order to (1) simplify the `hypertrain` target (now it does only one of two separate tasks: either creates a new sweep, or starts jobs), and (2) allow the var `$(SWEEP_REAL)` to be evaluated by `make` not by `bash`, so that the output is much more readable.
4. Run `wandb status` at the beginning of all hypertrain jobs to facilitate further debugging (it's a common error, to have an `uncategorized` wandb project when the sweep was created for a named project locally: `Sweep art-em/uncategorized/60jjg047 not found`)


ayushkovskiy:
Considering https://github.com/neuromation/cookiecutter-neuro-project/pull/338#issuecomment-580957475, we might want to save env var `RESULTS_DIR_STORAGE`. I'll re-make this PR.

ayushkovskiy:
Not needed: I found a simpler way of making template work on windows, without these changes needed.

ayushkovskiy:
After researching the problem, I found out that there's no stable way to run `make` from shell (which in turn runs by `make`) on Windows:
1. if `make` was installed by `conda install -c conda-forge make`, then:
- `make` is available from all make targets:
```
.PHONY: target
target:
    make ps  # works
```
- but `make` is not available from any of shell `sh` or `bash` (or `$(shell ..)`), which is needed by `hypertrain` target:
```
.PHONY: target
target:
    echo "Start" && make ps && echo "Finish"  # does not work
```
2. if `make` was installed by `conda install -c anaconda make`, then:
- I observed the same problem @mariyadavydova observed:
```
(base) C:\Users\User\azure>make ps
      0 [main] make 9172 find_fast_cwd: WARNING: Couldn't compute FAST_CWD pointer.  Please report this problem to
the public mailing list cygwin@cygwin.com
neuro ps
ERROR: Config at C:\Users\User\Anaconda3\home\User\.neuro does not exists. Please login.
Makefile:506: recipe for target `ps' failed
make: *** [ps] Error 70
```
however `neuro` itself is configured:
```
(base) C:\Users\User\azure>neuro ps
ID  NAME  STATUS  WHEN  IMAGE  OWNER  CLUSTER  DESCRIPTION  COMMAND
```
3. Also, I explored another temporary solution where `hypertrain` target only generates a shell script and then runs it separately, but, again, there's no way to run a `.sh` script from Makefile on Windows.

--
Since we are going to move logic (loop) from shell to python, I suggest us to consider this temporary solution to avoid calling `make` from shell.

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -340,56 +332,57 @@ kill-train:  ### Terminate the training job (set up env var 'RUN' to specify the
...
-	done ; \
-	echo "Please remove unused sweeps from the file ${WANDB_SWEEPS_FILE}" || :
+	@[ ! -f "$(WANDB_SWEEPS_FILE)" ] && echo "File '$(WANDB_SWEEPS_FILE)' does not exist. Abort." || { \
+		for sweep in $$(tac $(WANDB_SWEEPS_FILE)); do \
+			jobs=$$(neuro --quiet ps --description="$(PROJECT_ID):train:$$sweep") && \
+			echo "Killing jobs of the sweep '$$sweep'..." && \
+			$(NEURO) kill $${jobs:-placeholder} ||: ; \
+		done; \
+		echo "Please remove unused sweeps from the file '$(WANDB_SWEEPS_FILE)'." ; \
+	}
```
ayushkovskiy: 1. added a message indicating that the file with sweeps does not exist.
2. removed unnecessary part `|| tail -r ${WANDB_SWEEPS_FILE}`

PR #343, opened 2020-01-31
[Makefile] Get rid of some unnecessary variables

ayushkovskiy:
1. Get rid of `$(NEURO)`.
1.1 IMO, we it's not the case when *the user* might need to re-configure neuro executable
1.2 Also, IMO we don't need it in our tests -- at least myself, I have never used additional information provided by `neuro --verbose --show-traceback`.
1.3 Even if this information was useful, it's a bad engineering decision to add functionality only to use it in tests
1.4 `$(NEURO)` looks much more bulky than simple `neuro`
2. Get rid of `$(PIP)` and `$(APT)`
2.1 Just because we never overload these variables. Why over-complicate Makefile?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 LINTER_DIRS := tests
-NEURO_COMMAND := "neuro --verbose --show-traceback"
 TMP_DIR := $(shell mktemp -d)
 
 .PHONY: init
@@ -44,11 +43,11 @@ test_unit:
 
 .PHONY: test_e2e_dev
 test_e2e_dev:
-	PRESET=cpu-small NEURO=$(NEURO_COMMAND)  pytest -s -v --environment=dev --tb=short tests/e2e
+	PRESET=cpu-small pytest -s -v --environment=dev --tb=short tests/e2e
 
 .PHONY: test_e2e_staging
 test_e2e_staging:
-	PRESET=gpu-small NEURO=$(NEURO_COMMAND)  pytest -s -v --environment=staging --tb=short tests/e2e
+	PRESET=gpu-small pytest -s -v --environment=staging --tb=short tests/e2e
 
 .PHONY: get_e2e_failures
 get_e2e_failures:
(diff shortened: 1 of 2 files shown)


mariyadavydova:
Disagree. I see more harm in removing these vars than in keeping them.

1. Having executable paths in vars is a Makefile style convention. If a person is familiar with Makefiles, then they will be comfortable with `$(NEURO)`, `$(APT)`, and `$(PIP)`. 
2. These vars may be a huge help when it comes to troubleshooting on the user side. Compare:
```
Please, run `make train NEURO='neuro --verbose --show-traceback'` 
and show us an output.
```
with
```
Please, go to Makefile, find a `train` target, find `neuro` there, 
replace it with `neuro --verbose --show-traceback`, 
run `make train`, show us an output, 
and don't forget to change it back to `neuro`.`
```
For me, the latter looks really weird.

Also, my position here is the following: if we use a particular tool, let's follow this tool conventions.

ayushkovskiy:
@mariyadavydova Please provide at least one valid case when one needs to change command for `neuro`, `apt` or `pip`.

I see only one of them: If something goes wrong, we might ask the user to run with `neuro --trace`. However, we have never seen this case yet. Nevertheless, we live with hardly readable file with tons of `$(NEURO)` now. 

So, I think, if the user *does need* to run their target with `neuro --verbose`, they will be able to replace `neuro` with `neuro --verbose`. However, in order to uderstand what `$(NEURO)` **in every single target* means, the user needs to go to Google (if they never used make before), find out the syntax of make's env vars, then *read the whole Makefile carefully*, find the place where the variable is defined, and change its value. And at this point the surprise comes: if one defines a make variable enclosed with quotes (which is valid for bash), make will replace this value exactly as it is, with quotes! So the user might run `"neuro --verbose" run ...`, which will fail. 

So I think we need to KISS Makefile as possible.

ayushkovskiy:
I see this PR as the first step in iterative Template simplification.

mariyadavydova:
> However, we have never seen this case yet.

We haven't seen a case of a user being unable to work with our Makefile yet as well. Our user base is not representative yet, as it is quite small so far.

Both of us try to solve _imaginary_ problems in this PR. Both of us have valid arguments proving this or another point of view. In such a situation I would prefer to avoid changing anything until we get more information. 

However, if this change is so valuable for you, OK, please, go for it. But remember that later it may become necessary to change it back.

mariyadavydova:
This PR is hanging around for a while. I'm still against this change. 

Let's implement other ideas which will help us to reduce the size and increase the readability of the Makefile, and then come back to this idea. Do you agree?

PR #344, opened 2020-02-03
[Makefile] Upgrade the template, reverting the user content

mariyadavydova:
Internal functionality, thus is not covered with tests.

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -496,3 +496,19 @@ lint:  ### Run static code analysis locally
 .PHONY: ps
 ps:  ### List all running and pending jobs
 	$(NEURO) ps
+
+.PHONY: _upgrade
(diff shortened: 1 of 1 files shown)


ayushkovskiy:
1. From my point of view, functionality of self-update should not be a part of Makefile itself, even if non-exposed to the user, as it will add more complexity to the Template upgrade problem. However, If solving this problem from the other point, via [a specific tool](https://pypi.org/project/cookiecutter-project-upgrader/), we better save user's choice to `config.yaml` (project name, project slug, etc.) and re-use it. Note, in current version "Project Name" is not specified in `cookiecutter` command (this is minor and super-rare case, but since this target exists, this is a bug).
2. Since, I hope, we are moving towards minimizing the size of auto-generated code, this target is temporary and is to be used only by @mariyadavydova , so if its presence solves the problem, I'm OK with it.

PR #368, opened 2020-03-24
[Makefile] Support job tags

yartem:


tests/e2e/test_e2e_template.py
```diff
@@ -604,7 +604,7 @@ def test_make_hypertrain(
     # Print wandb status for debugging reasons
     run("wandb status")
 
-    n = 2
+    n = 1
```
ayushkovskiy: `n = 2` flakes sometimes: while the second job is pending, the first completes the sweep and thus the second has nothing to run.

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -394,7 +397,7 @@ connect-train: _check_setup  ### Connect to the remote shell running on the trai
 jupyter: _check_setup upload-config upload-code upload-notebooks ### Run a job with Jupyter Notebook and open UI in the default browser
 	$(NEURO) run \
 		--name $(JUPYTER_JOB) \
-		--description "$(PROJECT_ID):jupyter" \
+		--tag "target:jupyter" --tag "jupyter-mode:$(JUPYTER_MODE)" $(_TAGS) \
```
yartem, line 400: Get rid of `jupyter-mode:` tag (see https://docs.google.com/document/d/1Mv6uqnlvA3FdjW2KCp2YMwC_oYKWviCrlTX-f9G0-f4/edit?disco=AAAAJX0ioDk)

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -484,7 +489,21 @@ lint:  ### Run static code analysis locally
...
+	$(NEURO) ps $(_TAGS)
+
+.PHONY: ps-train
+ps-train:  ### List all running and pending training jobs
+	$(NEURO) ps --tag "target:train" $(_TAGS)
+
+.PHONY: ps-hypertrain
+ps-hypertrain:  ### List running and pending jobs of the latest hyper-parameter search sweep
+	sweep=`tail -1 $(WANDB_SWEEPS_FILE)` && \
+	$(NEURO) ps --tag "target:hypertrain" --tag "sweep:$$sweep" $(_TAGS)
```
yartem, line 501: add also tag `"target:train"`

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -34,7 +37,8 @@ CUSTOM_ENV=image:neuromation-$(PROJECT):$(VERSION)
 # Allows to set the `neuro` executable:
 #   make setup NEURO=/usr/bin/neuro
 #   make setup NEURO="neuro --verbose --show-traceback"
-NEURO?=neuro
+NEURO=neuro
```
mariyadavydova, line 40: Just to clarify: if you set a variable with `=`, doesn't that mean that you won't be able to overwrite it with `make setup NEURO=/usr/bin/neuro`?

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -433,21 +443,17 @@ hypertrain: _check_setup wandb-check-auth   ### Run jobs in parallel for hyperpa
 	done; \
 	echo -e "\nStarted $(N_HYPERPARAM_JOBS) hyper-parameter search jobs of sweep '$$sweep'.\nUse 'neuro ps' and 'neuro status <job>' to check."
 
+.PHONY: kill-hypertrain
+kill-hypertrain:  ### Terminate hyper-parameter search training jobs of the latest sweep
+	sweep=`tail -1 $(WANDB_SWEEPS_FILE)` && \
+	jobs=`neuro -q ps --tag "target:hypertrain" --tag "sweep:$$sweep" $(_PROJECT_TAGS) | tr -d "\r"` && \
```
mariyadavydova, line 449: We also agreed to change `sweep` to `wandb-sweep`.

{{cookiecutter.project_slug}}/Makefile
yartem: OK good question. Here's the reference: https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting
- `=`, `:=`, or `::=` define variable always:
```
[ay@archlinux temp]$ cat Makefile 
VAR=default
target:
	@echo $(VAR)

[ay@archlinux temp]$ make target 
default
[ay@archlinux temp]$ make target VAR=value1
value1
[ay@archlinux temp]$ export VAR=value2
[ay@archlinux temp]$ make target
default
```
note the exported variable did not override the value set in Makefile.

- `If you’d like a variable to be set to a value only if it’s not already set, then you can use the shorthand operator ‘?=’ instead of ‘=’.`:
```
[ay@archlinux temp]$ unset VAR
[ay@archlinux temp]$ cat Makefile 
VAR?=default
target:
	@echo $(VAR)

[ay@archlinux temp]$ make target VAR=value3
value3
[ay@archlinux temp]$ make target 
value5
[ay@archlinux temp]$ make target VAR=value6
value6
```

So we mark `?=` variables that act as target arguments (`make train RUN=value`, `make train PRESET=cpu-small`), and we mark `=` variables that are not supposed to be inline-overriden by the user (`make train PROJECT_ID=value` is discouraged).

{{cookiecutter.project_slug}}/Makefile
yartem: oops sorry missed that

PR #370, opened 2020-03-25
[Template] Exclude cache files while uploading files on Storage

yartem:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/93


{{cookiecutter.project_slug}}/.neuro.toml
```diff
@@ -0,0 +1,2 @@
+[storage]
+cp-exclude = ["*.pyc", ".mypy_cache", "__pycache__", ".ipynb_checkpoints", "wandb"]
```
mariyadavydova: Are you sure about 'wandb'? Won't this exclude affect hyperparameter tuning (as W&B probably requires some files to start new jobs during the tuning process)?

yartem: it contains logs "*.log" and settings. Removed from the list.

PR #372, opened 2020-03-26
[Makefile] Do not require setup for upload/download/clean targets

yartem:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/371

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -159,63 +159,63 @@ _check_setup:
 ##### STORAGE #####
 
 .PHONY: upload-code
-upload-code: _check_setup  ### Upload code directory to the platform storage
+upload-code:  ### Upload code directory to the platform storage
 	$(NEURO) cp --recursive --update --no-target-directory $(CODE_DIR) $(PROJECT_PATH_STORAGE)/$(CODE_DIR)
 
 .PHONY: download-code
-download-code: _check_setup  ### Download code directory from the platform storage
+download-code:  ### Download code directory from the platform storage
 	$(NEURO) cp --recursive --update --no-target-directory $(PROJECT_PATH_STORAGE)/$(CODE_DIR) $(CODE_DIR)
 
 .PHONY: clean-code
-clean-code: _check_setup  ### Delete code directory from the platform storage
+clean-code:  ### Delete code directory from the platform storage
 	$(NEURO) rm --recursive $(PROJECT_PATH_STORAGE)/$(CODE_DIR)/*
 
 .PHONY: upload-data
-upload-data: _check_setup  ### Upload data directory to the platform storage
+upload-data:  ### Upload data directory to the platform storage
 	$(NEURO) cp --recursive --update --no-target-directory $(DATA_DIR) $(DATA_DIR_STORAGE)
 
 .PHONY: download-data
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -159,63 +159,63 @@ _check_setup:
 ##### STORAGE #####
 
 .PHONY: upload-code
-upload-code: _check_setup  ### Upload code directory to the platform storage
+upload-code:  ### Upload code directory to the platform storage
```
mariyadavydova: There is a sense to keep _check_setup or to change the behavior of the commands somehow, because as it is now, upload won't work until you create a parent directory. I just checked on the latest version:

```
$ mkdir _test_
$ touch _test_/test.txt
$ neuro cp --recursive --update --no-target-directory _test_ storage:_test_1_/_test_2_
Copy 'file:///Users/mariyadavydova/projects/neuromation/_test_' => 'storage://neuro-ai-public/mrsmariyadavydova/_test_1_/_test_2_'
ERROR: cannot copy file:///Users/mariyadavydova/projects/neuromation/_test_ to storage://neuro-ai-public/mrsmariyadavydova/_test_1_/_test_2_: [Errno 2] No such directory: 'storage://neuro-ai-public/mrsmariyadavydova/_test_1_'
```

yartem: So you see the error and it's fine. This can be fixed with a small `neuro mkdir` command. From my understanding, `make setup` is used to *build image* which is a heavy operation not to be run often, therefore we came up to the `_setup_done` solution.

yartem: if following your logic, we'll need to make every single target depend on `_check_setup` since nothing will work as expected without the setup, even `make kill-*`

mariyadavydova: > This can be fixed with a small neuro mkdir command.

Users don't bother about small fixes (especially when they are new to a product). They see an error and close an app without even reading the error message.

> if following your logic, we'll need to make every single target depend on _check_setup since nothing will work as expected without the setup, even make kill-*

No, `make kill-*` should fail gracefully, because you don't always have a job. But yes, most of the targets (the ones expecting the particular directory structure) should depend on creating folders on storage.

I would suggest to make a small target, `_init_structure_`, which simply creates folders, make almost all targets (including `setup`) depending on it, and have the same `.init_done` file logic for it as we have for setup. What do you think?

yartem: > No, make kill-* should fail gracefully

Note: by design, killing a non-existing job will fail, killing a terminated (also killed) job will succeed. So we are moving towards "do nothing" on `make kill` if there's no such jobs running (will be implemented in https://github.com/neuromation/cookiecutter-neuro-project/pull/368)

yartem: > I would suggest to make a small target, _init_structure_ ...

True, it could be done like this. However I see it over-engineering since
1. error printed by `neuro cp` "Directory does not exist" is self-contained and it's clear that one needs to create a directory.
2. in fact we don't have a use-case or complain about hte problem

So there's no need to split `make setup` into two targets.

yartem:
Closing as non-necessary PR.

PR #373, opened 2020-03-26
[Makefile] Add commands to live-stream job's output

yartem:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/364


yartem:
> However, this change may become outdated pretty soon

OK, will not over-complicate and close this.

PR #379, opened 2020-03-27
[Makefile] Add argument RUN_EXTRA to all targets that run a job

yartem:
closes https://github.com/neuromation/cookiecutter-neuro-project/issues/227

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -1,93 +1,118 @@
-##### PATHS #####
-
 VERSION=v1.4
 BASE_ENV_VERSION=v1.4
 PROJECT_ID=placeholder
 
-DATA_DIR?=data
-CONFIG_DIR?=config
-CODE_DIR?={{cookiecutter.code_directory}}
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -1,111 +1,136 @@
...
 
-# The type of the training machine (run `neuro config show` to see the list of available types).
+# The type of the training machine (run `neuro config show`
+# to see the list of available types):
+#   make jupyter PRESET=cpu-small
 PRESET?=gpu-small
 
+# Extra options for `neuro run` targets:
+#   make train RUN_EXTRA="--env MYVAR=value"
+RUN_EXTRA?=
```
mariyadavydova: Can you please provide more examples of when this feature may be useful?

I see an example with TensorFlow logs in the issue, and it seems to be a rare case (I may be wrong here, of course).

yartem: this is an argument to run a job with some custom extra arguments without modifying Makefile:
`make train RUN_EXTRA="--env MYVAR=value --volume storage:mine:/mine:ro"`

PR #380, opened 2020-03-27
[Makefile] Add trailing slashes to directories

yartem:
To improve readability add trailing slashes to directories

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -35,7 +35,7 @@ JUPYTER_MODE?=notebook
 
 # Location of your dataset on the platform storage. Example:
 # DATA_DIR_STORAGE?=storage:datasets/cifar10
-DATA_DIR_STORAGE?=$(PROJECT_PATH_STORAGE)/$(DATA_DIR)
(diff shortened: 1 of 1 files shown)


yartem:
Sure, they were missing! fixed.

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -35,7 +35,7 @@ JUPYTER_MODE?=notebook
 
 # Location of your dataset on the platform storage. Example:
 # DATA_DIR_STORAGE?=storage:datasets/cifar10
-DATA_DIR_STORAGE?=$(PROJECT_PATH_STORAGE)/$(DATA_DIR)
+DATA_DIR_STORAGE?=$(PROJECT_PATH_STORAGE)/$(DATA_DIR)/
```
mariyadavydova: Looks like this '/' is redundant, because later you use `$(DATA_DIR_STORAGE)/`.

Also, some '/' are still missing.

Overall, I think that this PR makes Makefile more complex. Without these trailing slashes, one could update paths to `/dir` or `/dir/`, but now one has to set all paths without trailing slashes, because otherwise they will get `//` in paths.

Can you please explain how this change is expected to make Makefile more readable?

yartem: okay , will close.

PR #381, opened 2020-03-27
[Makefile] Upload/download only notebooks, improve readability

yartem:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/172 and https://github.com/neuromation/cookiecutter-neuro-project/issues/110

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -160,11 +160,19 @@ _check_setup:
 
 .PHONY: upload-code
 upload-code: _check_setup  ### Upload code directory to the platform storage
-	$(NEURO) cp --recursive --update --no-target-directory $(CODE_DIR) $(PROJECT_PATH_STORAGE)/$(CODE_DIR)
+    $(NEURO) cp \
+		--recursive \
+		--update \
+		--no-target-directory \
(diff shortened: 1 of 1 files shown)


ayushkovskiy:
Also closes https://github.com/neuromation/cookiecutter-neuro-project/issues/110

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -127,9 +127,9 @@ setup: ### Setup remote environment
 	$(NEURO) mkdir --parents $(PROJECT_PATH_STORAGE) \
 		$(PROJECT_PATH_STORAGE)/$(CODE_DIR) \
 		$(DATA_DIR_STORAGE) \
-		$(PROJECT_PATH_STORAGE)/$(CONFIG_DIR) \
+		$(PROJECT_PATH_STORAGE)/$(CONFIG_DIR)/ \
```
mariyadavydova: Can you please clarify, why most of the paths on the storage and in the container have now '/' at the end, but at the same time some of them don't?

mariyadavydova: Oh, I see now, that this PR is built on top of #380. So the change here seems to be fine.

PR #382, opened 2020-03-27
[Makefile] Add target: make save-develop

yartem:
Closes https://github.com/neuromation/cookiecutter-neuro-project/issues/256

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -27,6 +27,7 @@ FILEBROWSER_JOB?=filebrowser-$(PROJECT)
 
 BASE_ENV_NAME?=neuromation/base:$(BASE_ENV_VERSION)
 CUSTOM_ENV_NAME?=image:neuromation-$(PROJECT):$(VERSION)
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -27,6 +27,7 @@ FILEBROWSER_JOB?=filebrowser-$(PROJECT)
 
 BASE_ENV_NAME?=neuromation/base:$(BASE_ENV_VERSION)
 CUSTOM_ENV_NAME?=image:neuromation-$(PROJECT):$(VERSION)
+DEV_ENV?=image:neuromation-$(PROJECT):$(VERSION)-dev
```
mariyadavydova: Such an update won't actually do anything, as you only save a job to an image with a new name and never ever use it.

Could you please explain the reasons behind this issue?

yartem: Otherwise I don't see to have a separate development target, it one can't save it for the future. You're right, we don't do anything with `DEV_ENV` image after creating it. Here my motivation was to show users a way to work with their "remote local workstation" -- development job. Not to loose their work they could create an image from the job.

yartem:
Closing this PR as unnecessary.

PR #390, opened 2020-04-29
[Makefile] Stabilize W&B Hypertrain + Set up CI for MacOS

yartem:
Problems:
1. some macs don't have `/proc/self/fd/2`
2. some `wandb` printed `Create sweep with ID: ...` instead of `Created sweep with ID: ...` so grep was failing
3. removed `grep` in piped operation since even if it failed, the pipe execution was continuing to `awk`

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -412,7 +412,8 @@ hypertrain: _check_setup wandb-check-auth   ### Run jobs in parallel for hyperpa
 	@[ -f "$(WANDB_SWEEP_CONFIG_PATH)" ] \
 		&& echo "Using W&B sweep file: ./$(WANDB_SWEEP_CONFIG_PATH)" \
 		|| { echo "ERROR: W&B sweep config file not found: '$$PWD/$(WANDB_SWEEP_CONFIG_PATH)'" >&2; false; }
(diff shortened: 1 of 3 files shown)


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -412,7 +412,7 @@ hypertrain: _check_setup wandb-check-auth   ### Run jobs in parallel for hyperpa
 	@[ -f "$(WANDB_SWEEP_CONFIG_PATH)" ] \
 		&& echo "Using W&B sweep file: ./$(WANDB_SWEEP_CONFIG_PATH)" \
 		|| { echo "ERROR: W&B sweep config file not found: '$$PWD/$(WANDB_SWEEP_CONFIG_PATH)'" >&2; false; }
-	WANDB_PROJECT=$(PROJECT) wandb sweep $(WANDB_SWEEP_CONFIG_PATH) 2>&1 | tee /proc/self/fd/2 | grep 'Created sweep with ID: ' | awk 'NF{ print $$NF }' >> $(WANDB_SWEEPS_FILE)
+	WANDB_PROJECT=$(PROJECT) wandb sweep $(WANDB_SWEEP_CONFIG_PATH) 2>&1 | tee -a $(WANDB_SWEEPS_FILE).log | grep 'Created sweep with ID: ' | awk 'NF{ print $$NF }' >> $(WANDB_SWEEPS_FILE)
```
yartem, line 415: - `/proc/stderr` does not exist in windows environment
- `/proc/self` does not exist on some Macs
So we'll dump wandb initialization messages to a local log file.

yartem:
Problems arose while working on https://github.com/neuromation/ml-recipe-hyper-search/pull/2

PR #409, opened 2020-08-01
Rename the Makefile target "init" to "setup"

serhiy-storchaka:


anayden:
@serhiy-storchaka I'm curious why we need to keep the `init` target (2f523f43090667e62260b99d82c3410ead45ee29)

serhiy-storchaka:
It is just for the case if somebody uses `make init` in they scripts or fingers' memory. See https://github.com/neuromation/neuro-flow/pull/18#pullrequestreview-459602678.

anayden (approved):
LGTM

PR #411, opened 2020-08-12
Drop W&B completely

yartem:


{{cookiecutter.project_slug}}/.gitignore
```diff
@@ -5,7 +5,6 @@
 /config/*.json
 /config/*.yml
 /.setup_done
-/.wandb_sweeps
```
YevheniiSemendiak, line 8: Do [this](https://github.com/neuromation/cookiecutter-neuro-project/blob/ay/drop-wandb/%7B%7Bcookiecutter.project_slug%7D%7D/.gitignore#L15) guy is kept for user *wandb* usage in perspective?

Do we need [this](https://github.com/neuromation/cookiecutter-neuro-project/blob/ay/drop-wandb/.gitignore#L16) one?

mariyadavydova: I agree that we should remove both.

anayden:
Removed all mentions of W&B mentioned by @YevheniiSemendiak

{{cookiecutter.project_slug}}/.gitignore
YevheniiSemendiak: ok, got it

YevheniiSemendiak:
lgtm, thank you, @anayden

YevheniiSemendiak (approved):
approve from my side

PR #417, opened 2020-08-31
[docs] Update template version, reflect it in RELEASE.md

YevheniiSemendiak:


yartem:
1. the change in Makefile `VERSION` -> `TEMPLATE_VERSION` is unnecessary, if we drop it, we can also drop the tag `[Makefile]`
2. the tests may have failed because of colored output (it adds specific symbols that broke regexps):
`AssertionError: not found job status in output: `Job: job-12d880a1-6b93-4f5e-b9e8-32adbbd64fc5`
Though it's weird that tonight's tests are green (they also used the latest `neuro` and `neuro` wasn't released since today 01:00 am). So another idea is that the tests broke on `dev`.

yartem:
also, tensorboard's job returned `curl: (22) The requested URL returned error: 503 \r\n`

yartem:
Oh I see, it's not the color who broke the regex, it's the new status!
```
1mStatus[0m: [33munknown[0m
```
hmm, what does it mean? 

Anyways , we definitely need to add `unknown` to the regex here: https://github.com/neuromation/cookiecutter-neuro-project/blob/b71946094c853af75e2bdd8147c5ac433d610b30/tests/e2e/helpers/runners.py#L400

mariyadavydova:
Most likely, it is "Cancelled" which is not yet fully deployed (a new status for the case when you terminate a job manually).

YevheniiSemendiak:
> Most likely, it is "Cancelled" which is not yet fully deployed (a new status for the case when you terminate a job manually).

As I see from the [code](https://github.com/neuromation/platform-client-python/blob/master/neuromation/api/jobs.py#L83), this status should not be sent by the API under normal conditions.
The thing is that these tests were run on `dev` cluster, which may be a playground for the development process and result in such errs.

Therefore, @yartem I would suggest not to add this status to regex, but later update it with the"Cancelled".

Also, may we consider running [all](https://github.com/neuromation/cookiecutter-neuro-project/blob/master/.azure-pipelines/stage-e2e.yml#L34) tests on `staging` ?

YevheniiSemendiak:
@yartem, as we discussed privately, the changes in parser should not be propagated into e2e tests, since the new job status (`Cancelled`) is not released yet.

I've tried to re-run e2e's in the public cluster (thx, @yartem for the hint) - all was OK.
Please, check whether all is OK for you, and we could proceed with a merge.

YevheniiSemendiak:
Tests and docs will be updated in #419, closing this PR.

PR #418, opened 2020-09-07
[Tests] Support new job status in tests

YevheniiSemendiak:
Since recently a job status `cancelled` have been added to killed jobs, we should update our tests.


yartem (approved):
thanks! :tada:

PR #419, opened 2020-09-08
[Template] Migrate template from makefile to neuro-flow

YevheniiSemendiak:
## Intent
Since [neuro-flow](https://github.com/neuromation/neuro-flow) project functionality grew dramatically, in this PR I would like to replace the functionality of the project template control script ([makefile](https://github.com/neuromation/cookiecutter-neuro-project/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/Makefile)) by the neuro-flow `live` mode of execution. 

## Workflow:
- [x] implement the functionality of the makefile using neuro-flow, publish draft PR (@YevheniiSemendiak )
- [x] brief review then fixes ( @mariyadavydova , @yartem then @YevheniiSemendiak )
- [x] analyze tests, perform minor refactoring if possible (@yartem , @YevheniiSemendiak )
- [x] rewrite those tests that should be rewritten (@YevheniiSemendiak @yartem ) 
- [ ] update docs (both readmes, help.md, release notes, https://docs.neu.ro) (@YevheniiSemendiak , @mariyadavydova )
- [x] final review (@mariyadavydova , @yartem , @dalazx )
- [ ] release new template, highlighting its non-compatibility with previous versions (


cookiecutter.json
```diff
@@ -1,5 +1,5 @@
 {
     "project_name": "Neuro Project",
-    "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '-').replace('_', '-') }}",
-    "code_directory": "modules"
+    "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '_').replace('-', '_') }}",
+    "code_directory": "code"
```
yartem, line 4: I remember Bender was against `/project/code` because there was a pip package `code`, but I [can't find it](https://pypi.org/search/?q=code).
I vote for `code` if there's no collision, or C-like naming `src`. But not `modules` because it's similar to `models`.

hooks/post_gen_project.py
```diff
@@ -1,22 +1,15 @@
...
+project_kebab = "{{ cookiecutter.project_slug }}".replace("_", "-")
+project_id = f"{project_kebab}-{uuid.uuid4().hex[:8]}"
 
-makefile_path = Path("Makefile")
+live_file_path = Path(".neuro/live.yml")
 
-text = makefile_path.read_text()
-text = re.sub(r"PROJECT_ID=placeholder", f"PROJECT_ID={project_id}", text)
+text = live_file_path.read_text()
+text = re.sub(r"project:project-kebab-name-placeholder", f"project:{project_kebab}", text)
```
yartem, line 12: what's kebab? :sweat_smile:

{{cookiecutter.project_slug}}/.neuro.toml
```diff
@@ -1,2 +1,3 @@
 [storage]
-cp-exclude = ["*.pyc", ".mypy_cache", "__pycache__", ".ipynb_checkpoints"]
+cp-exclude = [".git", "*.pyc", ".mypy_cache", "__pycache__", ".ipynb_checkpoints", ".neuro"]
```
yartem, line 2: minor: let's just keep `[".git", ".neuro"]` because everything else is in gitignore.

YevheniiSemendiak: Goot catch, changed in 3524734

hooks/post_gen_project.py
YevheniiSemendiak: it a `kebab-case` 😄 (like a camelCase)
I just found out that the meaning is the same as `slug-case`, changed it back in 3524734.)

cookiecutter.json
YevheniiSemendiak: I also do like `src` more than modules or code. Thanks for the suggestion!
Changed in 3524734

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,55 @@
+name: CI
```
yartem, line 1: base CI config: neuro-extras config https://github.com/neuro-inc/neuro-extras/blob/master/.github/workflows/ci.yaml

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+
+  test:
+    name: Run tests
+    strategy:
+      matrix:
+        python-version: [3.7]
+        os: [ubuntu]
+        #python-version: [3.6, 3.7]
+        #os: [ubuntu, macos, windows]
```
yartem, line 21: no need to run units on different versions

{{cookiecutter.project_slug}}/.neuro/live.yml
```diff
@@ -0,0 +1,158 @@
+kind: live
+title: {{ cookiecutter.project_name }}
+id: PROJECT_SLUG_WITH_UNDERSCORES_PLACEHOLDER
+
+defaults:
+  preset: gpu-small-p
+  life_span: 1d
+  tags:
+    - kind:flow
+    - flow-id-with-dashes:{{ cookiecutter.project_slug }}
```
asvetlov, line 10: What is the purpose of these custom tags?

{{cookiecutter.project_slug}}/.neuro/live.yml
```diff
@@ -0,0 +1,158 @@
...
+
+volumes:
+  data:
+    remote: storage:$[[ flow.flow_id ]]/data
```
asvetlov, line 14: `$[[ flow.flow_id ]]` is calculated to `live` string here and below.
Do you really need `flow_id`? Maybe `project_id` fits better?

{{cookiecutter.project_slug}}/.neuro/live.yml
```diff
@@ -0,0 +1,158 @@
...
+    mount: /project/results
+    local: results
+  project:
+    remote: storage:$[[ flow.flow_id ]]
+    mount: /project
+    local: .
+
+images:
+  myimage:
+    ref: image:$[[ flow.flow_id ]]/develop
```
asvetlov, line 41: A constant `develop` tag looks like a source of errors.
I can imagine a calculated tag which is produced from hashes of Dockerfile, requirements, apt-requirement files, etc.
The approach may require something like not-existing-yet `hash_files()` function (see github counterpart https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles)
Adding the function to neuro-flow is not a hard task and can be an interesting exercise.

{{cookiecutter.project_slug}}/.neuro/live.yml
```diff
@@ -0,0 +1,158 @@
...
+    volumes:
+      - $[[ volumes.data.ref_ro ]]
+      - $[[ volumes.code.ref_ro ]]
+      - $[[ volumes.config.ref_ro ]]
+      - $[[ volumes.results.ref_rw ]]
+    env:
+      EXPOSE_SSH: "yes"
+      PYTHONPATH: $[[ volumes.code.mount ]]
+    bash: |
+        cd $[[ volumes.project.mount ]]
```
asvetlov, line 74: Please use `workdir` yaml option instead.

{{cookiecutter.project_slug}}/.neuro/live.yml
```diff
@@ -0,0 +1,158 @@
...
+      - $[[ volumes.results.ref_rw ]]
+    env:
+      EXPOSE_SSH: "yes"
+      PYTHONPATH: $[[ volumes.code.mount ]]
+    multi: true
+    bash: |
+        cd $[[ volumes.project.mount ]]
+        python $[[ volumes.code.mount ]]/train.py --data $[[ volumes.data.mount ]] $[[ multi.args ]]
+
+  jupyter:
```
asvetlov, line 94: I believe the next neuro-flow version will support actions-from-github; `jupyter`, `tensorboard` and `filebrowser` items will be replaced with action calls.

{{cookiecutter.project_slug}}/.neuro/live.yml
```diff
@@ -0,0 +1,158 @@
...
+      PYTHONPATH: $[[ volumes.code.mount ]]
+    cmd: >-
+      jupyter notebook
+        --no-browser
+        --ip=0.0.0.0
+        --allow-root
+        --NotebookApp.token=
+        --notebook-dir=$[[ volumes.notebooks.mount ]]
+
+  jupyter_lab:
```
asvetlov, line 116: What is the point of placing both `notebook` and `lab` modes in the same file?
Do people use both tools at the same time or select a single tool for the project?

{{cookiecutter.project_slug}}/HELP.md
```diff
@@ -47,132 +47,7 @@ make jupyter
...
-## Data
-
-### Uploading to the Storage via Web UI
-
-On local machine, run `make filebrowser` and open the job's URL on your mobile device or desktop.
-Through a simple file explorer interface, you can upload test images and perform file operations.
-
-### Uploading to the Storage via CLI
-
-On local machine, run `make upload-data`. This command pushes local files stored in `./data`
```
YevheniiSemendiak, line 72: mb lets leave these hints too?
just replacing `make` to `neuro-flow upload data` to upload **data** directory to the neuro storage.

{{cookiecutter.project_slug}}/HELP.md
```diff
@@ -47,132 +47,7 @@ make jupyter
...
-
-Please refer to the [documentation](https://docs.neu.ro/toolbox/accessing-object-storage-in-gcp) explaining how to start using GCP with the template.
-
-### Uploading data to the Job from AWS S3
-
-AWS CLI is pre-installed on all jobs produced from the Base Image.
-
-Please refer to the [documentation](https://docs.neu.ro/toolbox/accessing-object-storage-in-aws) explaining how to start using AWS with the template.
-
-## Customization
```
YevheniiSemendiak, line 87: Maybe let's leave this section, but just changing pointers onto `/{{ cookiecutter.project_slug }}/.neuro/live.yml` file?
Since this functionality (e.g. python debugging, updating data manually) is preserved (except Weights & Biases) and may be helpful.

{{cookiecutter.project_slug}}/README.md
```diff
@@ -4,14 +4,15 @@ The project description is here.
...
+Sign up at [neu.ro](https://neu.ro) and setup your local machine according to [instructions](https://docs.neu.ro/).
  
 Then run:
 
 ```shell
+pip install -U neuromation
 neuro login
-make setup
-make jupyter
+neuro-flow build {{ cookiecutter.project_slug }}
```
YevheniiSemendiak, line 14: I liked the idea of keeping the image name similar to the project slug, but if you feel that it is better to keep it as `myimage`, let's change it also here.
```suggestion
neuro-flow build myimage
```

{{cookiecutter.project_slug}}/README.md
```diff
@@ -4,14 +4,15 @@ The project description is here.
 
 ## Quick Start
 
-Sign up at [neu.ro](https://neu.ro) and setup your local machine according to [instructions](https://neu.ro/docs).
+Sign up at [neu.ro](https://neu.ro) and setup your local machine according to [instructions](https://docs.neu.ro/).
  
 Then run:
 
 ```shell
+pip install -U neuromation
```
YevheniiSemendiak, line 12:
```suggestion
pip install -U neuromation neuro-flow
```

YevheniiSemendiak (commented):
@yartem, thank you a lot for such a great job, I like the structure of the test and all other enhancements 👍 👍 👍 
Below I've left several comments, lets discuss them 🙂

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: well, we want to include flow-id here, but we can't use `flow-id` as it contains underscores what are forbidden in job tags.

I'm ready to drop this tag at all.

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: right! thanks

yartem: actually, you can see in tests `test_live_yml.py` that `flow_id` is evaluated to `my_project`

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: ok I'm changing this to `ref: image:$[[ flow.flow_id ]]`. Note, we use implicit `:latest` here, should we fixate a specific version? 

cc @mariyadavydova

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: ok fixed to `cd $[[ flow.workspace ]]`

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: great! we'll keep these here for now, and drop them after nf is released.

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: can we run sth like `neuro-flow run jupyter --mode=lab` ?

cc @mariyadavydova

hooks/pre_gen_project.py
```diff
@@ -1,23 +1,22 @@
 import re
 import sys
 
-PROJECT_REGEX = r'^[-a-zA-Z][-a-zA-Z0-9]+$'
+SLUG_REGEX = r'^[a-zA-Z][-a-zA-Z0-9]+$'
```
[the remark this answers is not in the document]
yartem: no, in this version we keep project-slug having `-` not `_`. Maybe later we'll change it

setup.cfg
```diff
@@ -28,14 +28,8 @@ incremental = False
 [mypy-pytest]
 ignore_missing_imports = true
 
-[mypy-pexpect]
```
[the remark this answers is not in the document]
yartem: it was for pip package `pexpect` we used instead of subprocess to test e2e

{{cookiecutter.project_slug}}/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM neuromation/base:v1.7.6
+
+ENV DEBIAN_FRONTEND noninteractive
+RUN mkdir /project
+WORKDIR /project
+
+COPY apt.txt .
+RUN export DEBIAN_FRONTEND=noninteractive && \
```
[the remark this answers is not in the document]
yartem: right, but it's better not to set this var permanently:
https://docs.docker.com/engine/faq/#why-is-debian_frontendnoninteractive-discouraged-in-dockerfiles

I'll remove line 3, thanks

{{cookiecutter.project_slug}}/HELP.md
yartem: ehhh I realized that this doc needs complete re-thinking, especially considering the fact that so far we didn't stabilize the template (there are still many things we want to drop/modify). So first, let's delete it.

{{cookiecutter.project_slug}}/README.md
yartem: right! thanks

{{cookiecutter.project_slug}}/README.md
yartem: exactly! :+1:

{{cookiecutter.project_slug}}/.neuro/live.yml
YevheniiSemendiak: I think yes, but the task should contain `multi:true` and we should run
`neuro-flow run jupyter -- --mode=lab`
those `--` indicate the start of *multi* parameters.

{{cookiecutter.project_slug}}/.neuro/live.yml
asvetlov: neuro-flow uses system tags to pin project, flow, and job ids to the executed job.
They are named like `project:{...}`, `flow:{...}` and `job:{...}`.
I suggest just keeping them

{{cookiecutter.project_slug}}/.neuro/live.yml
asvetlov: The test is not correct I guess.
Put `echo ${{flow.flow_id}}` for any job in live mode and execute it. You'll see `live` printout.

tests/unit/test_live_yml.py
```diff
@@ -0,0 +1,231 @@
+import textwrap
+from datetime import timedelta
+from typing import Awaitable, Callable
+
+import pytest
+from neuro_flow.context import JobCtx
+from neuro_flow.live_runner import LiveRunner
```
asvetlov, line 7: The `neuro-flow` tool has no API at all, you should never import something from the `neuro_flow` package.
Get ready to tests failures caused by neuro-flow internals changing. :)

setup.cfg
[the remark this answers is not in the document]
YevheniiSemendiak: you mean `[mypy-pexpect]`, or entire `setup.cfg` file?

{{cookiecutter.project_slug}}/HELP.md
YevheniiSemendiak: [resolving](https://github.com/neuro-inc/cookiecutter-neuro-project/pull/419#discussion_r498289193)

tests/unit/test_live_yml.py
yartem: I though we're supposed to be able to use neuro-flow's Python API, here's the first use-case.

setup.cfg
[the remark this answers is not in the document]
yartem: mypy-pexpect

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: :+1: removing it at all

{{cookiecutter.project_slug}}/.neuro/live.yml
```diff
@@ -0,0 +1,158 @@
+kind: live
+title: {{ cookiecutter.project_name }}
+id: PROJECT_SLUG_WITH_UNDERSCORES_PLACEHOLDER
```
yartem, line 3: note: neuro-flow doesn't support slug with `-`, only with `_`. So even if the directory (slug) is named with `-`, the dir on sotrage will be named with `_`. We should fix it.

tests/unit/test_live_yml.py
asvetlov: No. The project has no api, like the `pip`. 
It is a command-line tool only.

.github/workflows/ci.yaml
yartem: uncommented for e2e

tests/unit/test_live_yml.py
yartem: very sad.

{{cookiecutter.project_slug}}/.neuro/live.yml
mariyadavydova: I think that we can simply create an action for Jupyter/JupyterLab with a parameter `mode`. I believe that people either use Jupyter Notebooks or JupyterLab inside one project.

{{cookiecutter.project_slug}}/.neuro/live.yml
```diff
@@ -0,0 +1,157 @@
...
+
+defaults:
+  preset: gpu-small-p
+  life_span: 1d
+  tags:
+    - kind:flow
```
mariyadavydova, line 9: Do we really need this one? Flow jobs already have many tags, containing the project id and kind of jobs (live).

mariyadavydova (changes requested):
The PR is in pretty good shape, thank you, @yartem and @YevheniiSemendiak!

There some small things I'd ask to fix, though (see my comments).

{{cookiecutter.project_slug}}/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+    apt-get -qq clean && \
+    apt-get autoremove && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY setup.cfg .
+
+COPY requirements.txt .
+RUN pip install --progress-bar=off -U --no-cache-dir -r requirements.txt
+
+RUN ssh-keygen -f /id_rsa -t rsa -N neuromation -q
```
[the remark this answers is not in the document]
yartem: to make giithub keys work, I guess 

https://github.com/neuro-inc/cookiecutter-neuro-project/blob/94e218b50b9b0bb9c796fdbf36665b02f80acc17/%7B%7Bcookiecutter.project_slug%7D%7D/Makefile#L123

it was added together with NNI integration ^

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: Added:
```
$ neuro-flow run jupyter -- notebook
$ neuro-flow run jupyter -- lab
```
AFAIK defaults are not supported.

{{cookiecutter.project_slug}}/.neuro/live.yml
yartem: removed at all

mariyadavydova (approved):
👍

mariyadavydova:
I merge this PR as a huge first step in updating the template. The next steps are:

* Update Help.md 
* Move Jupyter etc in public actions 
* Draft the user docs update
* Test the template manually 
* Release both template and docs

PR #420, opened 2020-09-11
Trial to fix flaky tests

YevheniiSemendiak:
1. Try disabling searchwindowsize (will search the entire buffer by default) at feff9ef .


yartem:
not sure if this really fixes, but [sounds reasonable](https://pexpect.readthedocs.io/en/stable/api/pexpect.html):
```
When the keyword argument searchwindowsize is None (default), the full buffer is searched at each iteration of receiving incoming data.
```

Let's keep this in mind and maybe include this change to future tests refactoring.

artemlops:
outdated.

PR #432, opened 2021-01-20
Speed up image build

anayden:
Use `cpu-small` preset for image builds instead of `cpu-micro` by default


YevheniiSemendiak (approved):
Good idea!
At last, our CI is green here :D 
Thank you, Alex! 👍

PR #451, opened 2021-03-23
Fix windows tests

artemlops:


anayden (approved):
Let's proceed with this solution and revisit later

YevheniiSemendiak (approved):
Let's go 👍

PR #453, opened 2021-03-25
Update template for neu.ro

anayden:
1. Replace `neuromation` with `neuro-inc` / `neuro` where applicable
2. Make config files writeable and add them under source control
3. Add `v1` tag for the image

### DIFF
--- a/tests/cookiecutter.yaml
+++ b/tests/cookiecutter.yaml
@@ -1,5 +1,5 @@
 default_context:
     full_name: "Neu.ro Test"
-    email: "test@neur.ro"
+    email: "test@neu.ro"
     github_username: "neuro-inc"
     project_name: Test Project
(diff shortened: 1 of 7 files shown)


tests/cookiecutter.yaml
```diff
@@ -1,5 +1,5 @@
 default_context:
-    full_name: "Neuromation Test"
-    email: "test@neuromation.io"
-    github_username: "neuromation"
+    full_name: "Neu.ro Test"
+    email: "test@neur.ro"
```
artemlops, line 3: is `neur.ro` correct?

artemlops (approved):
:+1:

tests/cookiecutter.yaml
anayden: damn it! thank you :)

PR #497, opened 2021-08-19
Add live.yml reference hints 

YevheniiSemendiak:
### Summary of changes
1. Added `project.yml` file into the template
2. Added comments regarding the structure of both `live.yml` and `project.yml` files. Those comments might be removed during the initialization (a separate question asks whether to do this) like:
```
cookiecutter ~/work/projects/neuro/cookiecutter-neuro-project
project_name [Neuro Project]: 
project_dir [neuro project]: test-project
project_id [neuro_project]: awesome_project
code_directory [modules]: 
preserve Neuro Flow template hints [yes]:  <<<
```
3. Removed notation of `project_slug` since its intent is not clear. Instead, replaced it with explicit `project_dir` and `project_id`. Relaxed requirements to both of them, adapted pre_ and post_ generate hooks, added tests.
4. Added more e2e tests, luckily NF now has `--dry-run` to verify the correctness of the description file.
5. Added Python 3.8 and 3.9 to tests.

P.S. Sorry for the huge number of changes in a single PR.


hooks/post_gen_project.py
```diff
@@ -0,0 +1,41 @@
+import re
+import sys
+from pathlib import Path
+
+
+# >>> Optionally clearing comments comments
```
anayden, line 6: typo - comments comments.
same thing at the end of the file

hooks/post_gen_project.py
```diff
@@ -0,0 +1,41 @@
...
+    "true": True,
+    "no": False,
+    "n": False,
+    "false": False,
+}
+PRESERVE_HINTS_ANSWER = (
+    "{{ cookiecutter['preserve Neuro Flow template hints'] | lower }}"
+)
+if PRESERVE_HINTS_ANSWER not in PRESERVE_HINTS_VARIANS:
+    print(
```
anayden, line 23: maybe raise exception instead?

hooks/post_gen_project.py
```diff
@@ -0,0 +1,41 @@
...
+        f"ERROR: '{PRESERVE_HINTS_ANSWER}' is not a valid answer, "
+        f"please select one among [{', '.join(PRESERVE_HINTS_VARIANS)}]."
+    )
+    sys.exit(1)
+else:
+    if not PRESERVE_HINTS_VARIANS[PRESERVE_HINTS_ANSWER]:
+        for f_name in COMMENTS_STRUCTURE:
+            f_path = Path(f_name)
+            if not f_path.exists():
+                print(f"WARNING: {f_name} does not exist, skipping.")
```
anayden, line 33: for template user this message would be not very informative. maybe add something like "skipping comments removal from the fike"?

hooks/pre_gen_project.py
```diff
@@ -1,23 +1,46 @@
...
 
-if not re.match(PROJECT_REGEX, project_slug):
-    print('ERROR: %s is not a valid project slug. Slug can only contain letters, digits, '
-          'and dashes.' % project_slug)
+PROJECT_DIR = "{{ cookiecutter.project_dir }}"
+
+FULL_PROJECT_DIR = (Path(".") / PROJECT_DIR).resolve()
+FORBIDDEN_CHARS = r'<>:"\/|?*'
+
+forbidden_chars_in_dir_name = any(ch in PROJECT_DIR for ch in FORBIDDEN_CHARS)
```
anayden, line 11: just an idea: `... = set(PROJECT_DIR) & set(FORBIDDEN_CHARS)` - may be easier to read. your solution is fine too, feel free to keep it

hooks/pre_gen_project.py
```diff
@@ -1,23 +1,46 @@
...
+
+if forbidden_chars_in_dir_name:
+    print(f"ERROR: '{PROJECT_DIR}' contains forbidden chars ({FORBIDDEN_CHARS})")
```
anayden, line 14: same question re exception as above

hooks/pre_gen_project.py
```diff
@@ -1,23 +1,46 @@
...
+    try:
+        FULL_PROJECT_DIR.mkdir(parents=True)
+    except Exception:
+        print(
+            f"ERROR: '{FULL_PROJECT_DIR}' is not a valid project directory name "
+            "since OS cannot create it."
+        )
+        sys.exit(1)
+    else:
+        FULL_PROJECT_DIR.rmdir()
```
anayden, line 26: not sure i understand your intention here. could you please explain

tests/e2e/test_live.py
```diff
@@ -1,13 +1,36 @@
...
+@pytest.mark.parametrize("preserve_comments", ["yes", "no"])
+def test_neuro_flow_live(cookies: Cookies, preserve_comments: str) -> None:
+    result = cookies.bake(
+        extra_context={
+            "project_dir": "test-project",
+            "project_id": "awesome_project",
+            "preserve Neuro Flow template hints": preserve_comments,
+        }
+    )
+    print(result.project_path)
```
anayden, line 18: let's remove print call from test

tests/unit/test_bake_project.py
```diff
@@ -1,30 +1,102 @@
...
+        l_com_exists = l_com in live_file_content
+        p_com_exists = p_com in project_file_content
+        if preserve_comments == "yes":
+            assert l_com_exists, ".neuro/live.yml file does not contain comments"
+            assert p_com_exists, ".neuro/project.yml file does not contain comments"
+        elif preserve_comments == "no":
+            assert not l_com_exists, ".neuro/live.yml file contains comments"
+            assert not p_com_exists, ".neuro/project.yml file contains comments"
+        else:
+            raise RuntimeError("This should not happen.")
```
anayden, line 102: maybe `f"invalid value {preserve_comments} for preserve_comments. only yes and no are allowed"`

tests/unit/test_bake_project.py
```diff
@@ -1,30 +1,102 @@
...
+@pytest.mark.parametrize("preserve_comments", ["yes", "no"])
+def test_project_descr_with_comments(cookies: Cookies, preserve_comments: str) -> None:
+    result = cookies.bake(
+        extra_context={
+            "project_dir": "project-with-comments",
+            "preserve Neuro Flow template hints": preserve_comments,
+        }
+    )
+    assert result.exit_code == 0
+    l_com = "https://neu-ro.gitbook.io/neuro-flow/reference/live-workflow-syntax"
```
anayden, line 86: the string to check seems fragile (may change in the future and break otherwise valid tests). can we check for presence lines with comment sign instead?

{{cookiecutter.project_dir}}/.neuro/live.yml
```diff
@@ -0,0 +1,331 @@
...
+    #     descr: The name2 description
+    ## The parameters can be used in expressions for calculating other job attributes.
+    ## Note: 'params' could be also used in 'action' description.
+    # browse: true
+    ## Whether to open the job HTTP URI in a browser after the job startup. False by default.
+    # detach: true
+    ## By default, 'neuro-flow run <job-id>' keeps the terminal attached to the spawned job.
+    ## This can help with viewing the job's logs and running commands in its embedded bash session.
+    ## Enable the detach attribute to disable this behavior.
+    # entrypoint: sh -c "Echo $HOME"
```
anayden, line 201: Echo -> echo

anayden (approved):
That's an amazing set of improvements to the template! Nice job.

I like how detailed the documentation in the live.yaml is, although I can imagine some people would get scared by such a long read :) Hopefully our users are brave)

Anyway, I would suggest to let Jenia Gorlova proofread the documentation part when she's available, but other than that - it's a great PR.

I've got a few minor comments/questions (as usual), but nothing really serious.

Thanks a lot.

hooks/post_gen_project.py
YevheniiSemendiak: I was also wondering why not just an exception, but it's a requirement of [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/advanced/hooks.html#exit-with-an-appropriate-status), they handle exit codes and raised exception will add a traceback, which we will not able to hide.
the same says their example (however, implicitly :D)

hooks/pre_gen_project.py
YevheniiSemendiak: same answer - cookiecutter's requirement

hooks/post_gen_project.py
YevheniiSemendiak: thank you! fixed in c7d7b05

hooks/post_gen_project.py
YevheniiSemendiak: you are right, changed the msg in c7d7b05

hooks/pre_gen_project.py
YevheniiSemendiak: wow, that's a cool hack! thank you for the hint, applied in c7d7b05

hooks/pre_gen_project.py
YevheniiSemendiak: I was thinking that cookiecutter will raise an exception that the folder existrs. However I've tested that - all is ok, removed that else block in c7d7b05

tests/e2e/test_live.py
YevheniiSemendiak: good catch, rmved in c7d7b05

tests/unit/test_bake_project.py
YevheniiSemendiak: yeah :) that was more just for us as devs, but its not explicit, thanks for the hint, changed in c7d7b05

tests/unit/test_bake_project.py
YevheniiSemendiak: good idea, made in c7d7b05

{{cookiecutter.project_dir}}/.neuro/live.yml
YevheniiSemendiak: good catch, made in c7d7b05

PR #505, opened 2021-09-06
added workflow

ghost:


.github/workflows/update-mlflow.yml
```diff
@@ -0,0 +1,35 @@
+name: Update mlflow tags
```
YevheniiSemendiak, line 1:
```suggestion
name: Update action versions
```

Let's change the workflow file name in a similar way since now it launches a script that updates all actions versions.

.github/workflows/update-mlflow.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  schedule:
+  - cron: 0 0 * * *
+env:
+  PIP_EXTRA_INDEX_URL: ${{ format('https://{0}:{1}@{2}/{0}/{3}', secrets.DEVPI_USER,
+    secrets.DEVPI_PASS, secrets.DEVPI_HOST, secrets.DEVPI_INDEX) }}
```
YevheniiSemendiak, line 7: Why do we need it?

update_action.py
```diff
@@ -0,0 +1,42 @@
+"""
+  Scans directories for files that satisfy a specific mask and updates
+  the actions' tags.
+  python3 update_action.py --input “one/*.y*ml two/*.y*ml three/*.y*ml” --name mlflow
+"""
+
+import argparse
+import glob
+import re
+from github import Github
```
YevheniiSemendiak, line 10: We don't have this library in the environment, in which you are running this script.
`make setup` in the workflow script does install some dependencies, but not this one.
I would suggest replacing `make setup` with `pip install pygithub==version` [here](https://github.com/neuro-inc/cookiecutter-neuro-project/pull/505/files#diff-b5d2517dc5a22dbb782b9abd6520d8407ec39825213ced57e1bd690f9d15e3fbR18).
WDYT?

update_action.py
```diff
@@ -0,0 +1,42 @@
...
+from pathlib import Path
+
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument("-i", "--inputs", type=str, help="Input directories to scan")
+
+parser.add_argument("-n", "--name", type=str, help="Name of the action")
```
YevheniiSemendiak, line 18: Why not all actions?

We might detect the workflow file by its `kind:`, or any other neuro-flow specific attribute and the action by `gh:` or `github:` prefix.
Otherwise, if we have 5 actions, we will have 5 launches of this script. Not a critical problem, but as for me a single launch is more attractive and comfy.

.github/workflows/update-mlflow.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v3.10.1
+      with:
+        token: ${{ steps.generate_token.outputs.token }}
+        branch: update/image-tags
+        title: Auto-update image tags
+        commit-message: Auto-update image tags
+        body: |
+          Update versions of image in helm values
+          configs to latest version
```
YevheniiSemendiak, line 35: I'm not sure this content is correct. There is no helm, neither images.
Lets add relevant info:
- simple way: just state that we are updating action versions.
- fancy way: we could perform `::set-output::` in the Python script below and [use it](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs) in PR body.

update_action.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+# def glob_re(pattern, strings):
+#     return filter(re.compile(pattern).match, strings)
```
YevheniiSemendiak, line 22: redundant?

update_action.py
```diff
@@ -0,0 +1,42 @@
...
+
+args = parser.parse_args()
+inputs = args.inputs.split(" ")
+name = args.name
+g = Github()
+repo = g.get_repo(f"neuro-actions/{name}")
+tag = list(repo.get_releases())[-1].title
+# tag = '2.3.4'
+pattern = re.compile(
+    f"gh:neuro-actions/{name}" + r"@[a-zA-Z]{1}\d{1,2}\.\d{1,2}\.\d{1,3}"
```
YevheniiSemendiak, line 32: The scheme in action reference might be not only `gh:`, but also `github:`: [reference](https://neu-ro.gitbook.io/neuro-flow/reference/live-workflow-syntax#jobs-less-than-job-id-greater-than-action).
Let's support both schemes in the script.

Also, let's parameterize the owner name, `neuro-actions`, since some actions might be in the user's repo.

YevheniiSemendiak (changes requested):
Thank you for such great improvement! 💪 
Below, I've left several questions/suggestions, some of them are mandatory as for me, some are optional, but general suggestion - let's generalize the script to update all actions!

Thank you!

dmitrynikolayev:
Pls, review the branch

.github/workflows/update-mlflow.yml
YevheniiSemendiak: @dmitrynikolayev , this issue is still actual. 
If you're not agreeing with the suggestion - let's discuss it here.

update_action.py
YevheniiSemendiak: @dmitrynikolayev , the same here. What do you think about the cases, when the action is hosted under a different repository?

.github/workflows/update-mlflow.yml
YevheniiSemendiak: Hint: https://stackoverflow.com/a/59201610

update_actions.py
```diff
@@ -0,0 +1,59 @@
...
+
+parser.add_argument(
+    '-s', '--sources',
+    type=str, help='Source directories to scan'
+)
+
+args = parser.parse_args()
+sources = args.sources.split(' ')
+g = Github()
+pattern = r"\s+action:.*"
```
anayden, line 24: Maybe using a regexp to catch all params would be easier? Something along the lines of 

```python
pattern = r"\s+action:\s*(?P<svc>gh|github):(?P<org>[\w-]+)/(?P<repo>[\w-]+)@(?P<version>v[\d.]+)"
r = re.compile(pattern)

...
for file_path in glob.iglob(source):
  for line in open(file_path).readlines():
    match = r.match(line)
    if match:
       # use match['svc'], match['org'], match['repo'] and match['version']
```

update_actions.py
```diff
@@ -0,0 +1,59 @@
...
+            svc_bgn, svc_end = 0, slug.find(':')
+            rep_bgn, rep_end = slug.find(':') + 1, slug.find('/')
+            acn_bgn, acn_end = slug.find('/') + 1, slug.find('@')
+            tag_bgn, tag_end = slug.find('@') + 1, len(slug)
+            svc = slug[svc_bgn:svc_end]
+            rep = slug[rep_bgn:rep_end]
+            acn = slug[acn_bgn:acn_end]
+            tag = slug[tag_bgn:tag_end]
+            p = re.compile(slug)
+            try:
```
anayden, line 45: This `try` block kind of lacks comments on what is happening inside.

update_actions.py
```diff
@@ -0,0 +1,59 @@
...
+                rep_path = f"{rep}/{acn}"
+                rep_obj = g.get_repo(rep_path)
+                if tag not in cache:
+                    cache[tag] = list(rep_obj.get_releases())[-1].title
```
anayden, line 49: What if my action file references not a version, but `master`? E.g. `action: gh:neuro-actions/filebrowser@master`. Looks like the code will replace my master reference with latest release. That's not desirable. Let's make sure this code only affects versioned tags + checks repo has at least one release (otherwise the code `get_releases[-1]` raises an exception)

update_actions.py
```diff
@@ -0,0 +1,59 @@
...
+                tag_ = cache[tag]
+                if tag != tag_:
+                    n_changed += 1
+                    ss = re.sub(p, f'{svc}:{rep}/{acn}@{tag_}', s)
+                    with open(file_path, "w") as file:
+                        file.write(ss)
+            except:
+                continue
```
anayden, line 57: Would you mind at least logging exceptions here? Otherwise we'll have zero visibility into any issues with the code.

update_actions.py
```diff
@@ -0,0 +1,59 @@
...
+            rep = slug[rep_bgn:rep_end]
+            acn = slug[acn_bgn:acn_end]
+            tag = slug[tag_bgn:tag_end]
+            p = re.compile(slug)
+            try:
+                rep_path = f"{rep}/{acn}"
+                rep_obj = g.get_repo(rep_path)
+                if tag not in cache:
+                    cache[tag] = list(rep_obj.get_releases())[-1].title
+                tag_ = cache[tag]
```
anayden, line 50: Name with underscore looks confusing (as if it collides with an internal Python keyword). How about renaming this to a `latest_tag` for instance?

update_actions.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+import glob
+import argparse
+import re
+from pathlib import Path
+from github import Github
+
+parser = argparse.ArgumentParser()
+cache = {}
```
anayden, line 14: let's add type annotation here:

```python
cache: Dict[str, str] = {}
```

update_actions.py
```diff
@@ -0,0 +1,59 @@
...
+g = Github()
+pattern = r"\s+action:.*"
+
+for source in sources:
+    path = Path(source)
+    for file_path in glob.iglob(source):
+        n_actions = 0
+        n_changed = 0
+        with open(file_path) as file:
+            s = file.read()
```
anayden, line 32: Let's make variable names more self-explanatory. Please rename `s`, `p`, `ss`, `g` to a more readable names.

.github/workflows/update-actions.yml
```diff
@@ -0,0 +1,31 @@
...
+        private_key: ${{ secrets.BOT_PRIVATE_KEY }}
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v3.10.1
+      with:
+        token: ${{ steps.generate_token.outputs.token }}
+        branch: update/actions-tags
+        title: Auto-update actions tags
+        commit-message: Auto-update actions tags
+        body: |
+          Update version tagos of actions found in input folders
```
anayden, line 31: typo: `tagos` → `tags`

update_actions.py
```diff
@@ -0,0 +1,59 @@
...
+                tag_ = cache[tag]
+                if tag != tag_:
+                    n_changed += 1
+                    ss = re.sub(p, f'{svc}:{rep}/{acn}@{tag_}', s)
+                    with open(file_path, "w") as file:
+                        file.write(ss)
+            except:
+                continue
+            n_actions += 1
+        print(f'::set-output {n_changed}/{n_actions} in {file_path} actions were updated')
```
anayden, line 59: What does it mean? How is this used? What is the result of this line? I'm confused…

anayden (commented):
Nice job on creating the script, but I feel it needs some updates.

dmitrynikolayev:
Tried to improve. Check this.

update_actions.py
YevheniiSemendiak: done in f38b8c7

YevheniiSemendiak:
@anayden, pls check it out.
Those failed tests in Windows are related to https://github.com/neuro-inc/neuro-flow/issues/331 , you could ignore them.

anayden:
> Btw, what do you think, if we add this script and the workflow to the template itself?

I don't see much difference, so feel free to follow your heart :)

PR #537, opened 2021-12-01
Rename `myimage` to `train` and add project description

YevheniiSemendiak:


anayden (approved):

requirements/pipx.txt
```diff
@@ -0,0 +1,2 @@
+neuro-cli==21.11.4
```
asvetlov, line 1: Replace these packages with a single `neuro-all` maybe?

asvetlov (approved):
lgtm

PR #696, opened 2024-11-08
migrate from neuro to apolo

costefan:


costefan:
@YevheniiSemendiak please take a look on Slack conversation and 401 error for the user: CLIENT_TEST_E2E_USER_NAME
apolo_sdk.AuthorizationError: {"missing": [{"uri": "flow://default/NO_ORG/conquering-the-realms", "action": "write"}]}\n')

